PR #33, opened 2019-04-28
Update logo_annotation_labels.txt

teolemon:


data/ocr/logo_annotation_labels.txt
```diff
@@ -1,3 +1,5 @@
 label rouge||en:label-rouge
+label rouge vectoriel||en:label-rouge
```
raphael0202: @teolemon What is this "label rouge vectoriel" ?

teolemon: some less frequent variation I found in an OCR file

PR #35, opened 2019-04-28
Update regex_brands.txt

teolemon:


data/ocr/regex_brands.txt
```diff
@@ -48,11 +56,14 @@ Gerblé
 Grand Jury
 Gut & Gunstig
 Gutbio
+Häagen-Dazs||H[aä]agen[- ]Dazs
```
raphael0202: @teolemon the regex part must be in lowercase

teolemon: fixed

PR #71, opened 2019-07-04
Add automatic processing to results for a small number of label regexp

CloCkWeRX:
Treat this more as psuedocode; completely 100% untested locally.

Adds a 'trust' weighting to OCRRegexp instances, to allow some to be high trust, others to be low trust; and to decide if automatic processing is appropriate.

This is based on my subjective observations of the various labels only, open to stats to set more realistic limits.

### DIFF
--- a/robotoff/insights/ocr/dataclass.py
+++ b/robotoff/insights/ocr/dataclass.py
@@ -28,13 +28,15 @@ class OCRRegex:
                  lowercase: bool = False,
                  processing_func: Optional[Callable] = None,
                  priority: Optional[int] = None,
-                 notify: bool = False):
+                 notify: bool = False,
+                 trust: float = 1):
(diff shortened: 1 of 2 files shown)


raphael0202:
Good idea!
Do you see any advantage of using a scalar for the `trust` field over a boolean field? If not, a boolean seems simpler to me.
The field must be added to all label OCRRegex, and there are some changes to be done in the label importer in `robotoff/insights/importer.py`.

CloCkWeRX:
I'm not too fussed either way, I picked a scalar to let us set thresholds based on statistics - ie; this regexp is 90% accurate; that one is 73% accurate; which could be later derived from user responses to skip/approve/disapprove match... but at the end of the day, it boils down to "do we trust it enough: yes/no".

PR #72, opened 2019-07-06
Add a few more store brands

hangy:
Some Aldi, Edeka, Lidl, and Woolworth store brands.


teolemon:
@raphael0202 you might have missed this one by @hangy

PR #84, opened 2019-08-20
add brands taxonomy

teolemon:
Currently a few brands are whitelisted for cleaning product name for matching.
Some other brands are whitelisted for REGEX matching in OCR.

This is a longer, curated list of brands, that may evolve overtime, very much like other taxonomies


teolemon:
@raphael0202 for more scalable brand extraction

raphael0202:
@teolemon This feature has been developed. However, there are weird items in the brand taxonomy JSON (en:france, en:fromage, en:artisanal), that I can't find in brands.txt on github (https://github.com/openfoodfacts/openfoodfacts-server/blob/master/taxonomies/brands.txt). Do you know where those come from?

teolemon:
This is a taxonomy of oft used values for brands. I think we have to clean it.

raphael0202:
Done.

PR #89, opened 2019-11-17
Add common Australian items to stores & brands

CloCkWeRX:
For 69ae403; not sure if this is useful, or if #84 is just easier.

I haven't explored much around the common aliases from OCR, but happy to add if things like apostrophes are often mistaken.

### DIFF
--- a/data/ocr/regex_brands.txt
+++ b/data/ocr/regex_brands.txt
@@ -23,6 +23,7 @@ Carrefour Baby||carrefour [bg]aby
 Carrefour Bio
 Carrefour Discount
 Chabrior
+Coles
 Colruyt
 Conad
 Coop Betty Bossi
--- a/data/ocr/regex_brands.txt
+++ b/data/ocr/regex_brands.txt
@@ -1,14 +1,19 @@
+Aeroplane Jelly
 Albert Heijn
 Aldi
 Alesto
+Allen's
 Alipende
 Alnatura
 Alpenmark
 Alteza
+Arnott's
 Anna's Best
 Asda
 Auchan
 Barilla
+Beerenberg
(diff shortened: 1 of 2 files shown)


data/ocr/regex_brands.txt
```diff
@@ -23,13 +28,16 @@ Carrefour Baby||carrefour [bg]aby
 Carrefour Bio
 Carrefour Discount
 Chabrior
+Coles
```
teolemon: Coleslow ? @raphael0202

raphael0202: @teolemon it seems that Coles is indeed a brand: https://au.openfoodfacts.org/brand/coles

raphael0202:
Thank you for the PR, and sorry for not replying sonner.
For brands and stores, a 'notify' list allows to be informed when automatic annotations are made with these brands/stores on Slack (on the robotoff-alert channel): https://github.com/openfoodfacts/robotoff/blob/master/data/ocr/notify_brands.txt and https://github.com/openfoodfacts/robotoff/blob/master/data/ocr/notify_stores.txt.
Some brands/stores seem a bit problematic to me (Peters, Coles, Bega, Albert's,...), as these are short (or common) strings that can easily be found in OCR texts. Can you trigger notifications for those (and other that seem problematic to you?). You can delete all current notifications for brands and stores, we had enough time to check that there were no false positive on those.

CloCkWeRX:
I'll tweak things for the ambiguous ones shortly then

raphael0202:
@CloCkWeRX I've implemented https://github.com/openfoodfacts/robotoff/pull/84, and those australian brand are already in the taxonomy

PR #96, opened 2020-01-15
more packaging

teolemon:


data/ocr/packaging.txt
```diff
@@ -30,3 +30,48 @@ Pot en plastique||pot plastique
 Pot en verre||pot verre
 Quadrette en plastique||quadrette plastique
 Sachet en plastique||sachet plastique
+Tetra Pak||tetra pak
+Tetra Pak||tetrapak
```
raphael0202: What's the difference between Tetra Pak and Tetra Pack?

teolemon: no idea. spelling ?

teolemon: I think it's the same

PR #125, opened 2020-03-05
Pseudo code to extract kJ

teolemon:
Pseudo code to extract kJ when it is missing from the product


raphael0202:
Can you transform this PR into an issue?

PR #126, opened 2020-03-05
Create a fiber check

teolemon:
Create a fiber check  to let contributors fill missing fiber (pseudocode)


raphael0202:
Wouldn't this be more appropriate as a quality facet?

raphael0202:
Implemented.

PR #130, opened 2020-03-23
adding-active-contribors

teolemon:
adding-active-contribors


raphael0202:
In your script, the date doesn't change, it is the date of launch of the scheduler. I reimplemented it here: dabd8dafb0a8af87a49814556cb2bf1aaeb16bfc

PR #136, opened 2020-04-08
Location extraction using regex

alex-marty:


raphael0202:
The new tests you added are failing, can you take a look?

scripts/extract_location.py
```diff
@@ -0,0 +1,48 @@
+from pathlib import Path
```
raphael0202, line 1: You should be able to generate location insights using:
`python3 -m robotoff ocr.jsonl.gz -t location -o location.jsonl`

raphael0202: So I don't think we need this script

robotoff/insights/ocr/location.py
```diff
@@ -0,0 +1,243 @@
...
+            city_start (int): Start index of the city name match in `text`.
+            city_end (int): End index of the city name match in `text`.
+
+        Returns:
+            (str, int, int) or None: If the `City`'s postal code was found close to
+            the city name match, it is returned along with its start and end indices
+            in the text. If it was not found, returns None.
+        """
+        if not city.postal_code.isdigit():
+            raise ValueError(f"postal code contains non-digit characters: {city}")
```
raphael0202, line 212: a logger.error followed by a "return" may be more appropriate, as it would prevent the main process from crashing, in case one of the postal code is malformed.

raphael0202:
It's all good to me, merging. Thanks!

PR #145, opened 2020-04-24
Extend generate-ocr-insights CLI command

alex-marty:
This PR adds to the generate-ocr-insights command the following abilities:
* keep documents with empty insights instead of filtering them out, with the `--keep-empty` flag (need this to get complete stats when processing the full OCR file)
* read JSONL input from stdin instead of a file, by specifying '-' as source (need this to process the full OCR file as a stream without downloading it first)

Also adds documentation to the command.


robotoff/__main__.py
```diff
@@ -31,16 +32,45 @@ def predict_insight(ocr_url: str):
...
+        show_default=True,
+    )
+    def generate_ocr_insights(
+        source: str, insight_type: str, output: str, keep_empty: bool
+    ):
         from robotoff.cli import insights
         from robotoff.utils import get_logger
 
+        if source == "-":
+            source = sys.stdin
```
raphael0202, line 70: mypy error here: Incompatible types in assignment (expression has type "TextIO", variable has type "str")

robotoff/cli/insights.py
```diff
@@ -19,7 +19,9 @@
 from robotoff.utils import gzip_jsonl_iter, jsonl_iter
 
 
-def run_from_ocr_archive(input_: str, insight_type: str, output: Optional[str]):
+def run_from_ocr_archive(
+    input_: str, insight_type: str, output: Optional[str], keep_empty: bool = False
```
raphael0202, line 23: Can you update type for input_?

robotoff/insights/ocr/core.py
```diff
@@ -109,20 +109,32 @@ def get_source(image_name: str, json_path: str = None, barcode: str = None):
...
+    for line in stream:
+        json_data = json.loads(line)
+
+        if "content" in json_data:
+            source = json_data["source"].replace("//", "/")
+            yield source, json_data["content"]
+
+
+def ocr_iter(source: Union[str, TextIO]) -> Iterable[Tuple[Optional[str], Dict]]:
+    if not isinstance(source, str):
```
raphael0202, line 122: It seems good to me. However, there are many mypy errors in ocr_iter function, due to the multiple types associated with ocr_iter source parameter. I know some of these errors were already there before your change, but can you however take a look at it? If you don't have time it's alright.

PR #146, opened 2020-04-25
Extend location insight

alex-marty:


teolemon:
@raphael0202 @raphodn regarding postal addresses

PR #158, opened 2020-04-30
Add FSC code regex

segysego:


robotoff/insights/ocr/packager_code.py
```diff
@@ -29,13 +28,24 @@ def process_fr_emb_match(match) -> str:
...
+
 PACKAGER_CODE: Dict[str, OCRRegex] = {
     "fr_emb": OCRRegex(
         re.compile(r"emb ?(\d ?\d ?\d ?\d ?\d) ?([a-z])?(?![a-z0-9])"),
         field=OCRField.text_annotations,
         lowercase=True,
         processing_func=process_fr_emb_match,
     ),
+    "fsc": OCRRegex(
+        re.compile(r"FSC.? (C\d{6})"),
```
raphael0202, line 44: I would suggest r"FSC.? ?(C\d{6})", so that "FSC-C135435" (the code that was initially given by @teolemon as example) is correctly detected.

robotoff/insights/ocr/packager_code.py
```diff
@@ -29,13 +28,24 @@ def process_fr_emb_match(match) -> str:
...
+        field=OCRField.text_annotations,
```
raphael0202, line 45: filed=OCRField.full_text is better here, we should use it in other packager code regexes as well.

robotoff/insights/ocr/packager_code.py
```diff
@@ -29,13 +28,24 @@ def process_fr_emb_match(match) -> str:
...
+        lowercase=False,
```
raphael0202, line 46: any reason why not lowercasing? The pattern seems specific enough to me to allow lowercasing.

robotoff/insights/ocr/packager_code.py
segysego: @raphael0202  I tried with full_text and with full_text_contiguous and it does not work for code 20126803. Only text_annotations works for this code.

raphael0202: I see, indeed full_text is not always present. My bad.

robotoff/insights/ocr/packager_code.py
segysego: done!

robotoff/insights/ocr/packager_code.py
segysego: @raphael0202 @teolemon 
I updated the branch with the modifications

raphael0202:
Merged, thanks for the PR!

PR #234, opened 2020-08-13
Add a RGSEAA flashtext list

teolemon:
- Add a RGSEAA flashtext list
**TODO**
- [ ] See if a REGEX is not a better way
- [ ] Several combinations: "RGSEAA" OR "R.G.S.E.A.A" / "-" OR  ":" OR " "


raphael0202:
A regex would definitely be a better way to detect those!
I'm closing this PR as it's not the best way to go.

PR #301, opened 2020-12-07
Auto-select recycling images (closes #260)

### LINKED ISSUE #260: Auto-select recycling photos based on OCR
The new recycling photos photo field which is now live.
https://world.openfoodfacts.org/product/3560070820306/fingers-de-poulet-carrefour

It would be tremendously useful if we could auto-select the most recent photo where we detect recycling instructions using a few OCR regex, then letting contributors recrop it and OCR it.

### Pseudo code
``` 
for product in products:
    if "packaging-photo-to-be-selected" in product.states:
        for image in product.images:
            if "consignesdetri.fr" or "recycler" in image.ocr:
                select_image(product, packaging-photo, fr)
```

### List of triggers
```
"consignesdetri.fr"
"recycler"
"jeter"
"bouteille"
…
```
### Selecting an image for Recycling works like selecting an image for ingredients 
```
https://world.openfoodfacts.org/cgi/product_image_crop.pl?code=8901063017252&id=packaging_en&imgid=14&x1=0&y1=0&x2=0&y2=0&coordinates_image_size=full&angle=0&normalize=false&white_magic=false
https://world.openfoodfacts.org/state/packaging-photo-to-be-selected.json
```

Asteur:
Issue #260 
Background task for updating packaging images with recycling labels


robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+def check_image_in_pack(image, lang, pack_images):
+    """
+    Check if image has been already selected
+
+    :param image: imgid
+    :param lang: en/fr/...
+    :param pack_images: image ids [packaging_en, front_en]
+    :return:
+    """
+    if "packaging_{}".format(lang) in pack_images:
```
alex-marty, line 134: This condition could be removed, as it is implied by the `.get()` on next line.

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+    :return:
+    """
+    field = "packaging_{}".format(lang)
+
+    data = get_json_for_image(product.get("code"), image)
+    image_text = data.get("responses", [])
+    if image_text:
+        image_text = image_text[0].get("fullTextAnnotation", {}).get("text", "")
+
+        if any([1 if i in image_text.lower() else 0 for i in recycling_triggers]):
```
alex-marty, line 104: This could be better written as:

```suggestion
        if any(trigger in image_text.lower() for trigger in recycling_triggers):
```

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+            ru = None
```
alex-marty, line 105: What does the variable `ru` mean? Could it be renamed to something more meaningful?

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+            if field in pack_images.keys():
+                ru = unselect_image(ean, field, username, password)
+
+            r = reselect_image(ean, field, image, username, password)
```
alex-marty, line 109: Same here, could this variable have a more explicit name?

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+            if r:
+                if ru:
+                    logger.info(
+                        "Recycle image(changed): ",
+                        ean,
+                        field,
+                        pack_images[field].get("imgid"),
+                        "->",
+                        image,
+                    )
```
alex-marty, line 120: This logger call will crash as there is no % formatting in the message.

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+                else:
+                    logger.info("Recycle image(selected): ", ean, field, "->", image)
```
alex-marty, line 122: Same here.

alex-marty: Also, could the block nesting in this function be reduced a bit?

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+        "code": barcode,
+        "id": field_name,
+        "user_id": username,
+        "password": password,
+    }
+    try:
+        r = requests.post(url, data=data)
+        if r.ok:
+            return True
+    except Exception:
```
alex-marty, line 164: Same as above regarding exception type. Also, what does an error here mean? Are you sure you want to ignore them?

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+        logger.warn(
+            "Exception in unselect_image: barcode - {}, id - {}".format(
+                barcode, field_name
+            )
+        )
```
alex-marty, line 169: When logging messages should not be formatted directly, but rather use the syntax `logger.warning("Message %s %s", x, y)`.

Also `warn()` is deprecated in favor of `warning()` (https://docs.python.org/3/library/logging.html#logging.Logger.warning).

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+
+    :param barcode: ean
+    :param field_name: packaging_en
+    :param img_id: imgid=[0, 1, 2]
+    :param username:
+    :param password:
+    :return:
+    """
+
+    if img_id:
```
alex-marty, line 186: This function should not receive an invalid `img_id`, this condition should be handled by the caller.

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+            "code": barcode,
+            "imgid": img_id,
+            "id": field_name,
+            "user_id": username,
+            "password": password,
+        }
+        try:
+            r = requests.post(url, data=data)
+            if r.ok:
+                return True
```
alex-marty, line 198: This, together with the final `return False`, could be rewritten as:

```suggestion
            return r.ok
```

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+        except Exception:
```
alex-marty, line 199: Same here about the exception handling. Is it a good thing to ignore errors?

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+            logger.warn(
+                "Exception in reselect_image: barcode - {}, id - {}, imgid - {}".format(
+                    barcode, field_name, img_id
+                )
+            )
```
alex-marty, line 204: Same about log message formatting.

alex-marty (changes requested):

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
+from robotoff.robotoff.products import ProductDataset
+from robotoff.robotoff.insights.extraction import get_logger, requests
+from robotoff.robotoff.insights.ocr.core import get_json_for_image
+
+logger = get_logger(__name__)
+
+
+def update_recycling(username, password):
```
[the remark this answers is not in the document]
Asteur: Ok, if it really helps

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    Function to update "Recycle" image for the product
+    based on triggers
+
+    :param username:
+    :param password:
+    :return:
+    """
```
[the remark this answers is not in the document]
Asteur: No problems, we can even remove these lines, it was made to describe "what's going in function"

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+    }
+    # get products dataset
+    dataset = ProductDataset.load()
+
+    # iterate products
+    for product in dataset.stream().filter_nonempty_text_field("code"):
+
+        if "packaging-photo-to-be-selected" in product.get("states", ""):
+            images = get_images(product.get("code"))
+            if images:
```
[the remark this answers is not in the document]
Asteur: Not sure how it helps, but ok

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+        "fr": ["consignesdetri.fr", "recycler", "jeter", "bouteille"],
+    }
+    # get products dataset
+    dataset = ProductDataset.load()
+
+    # iterate products
+    for product in dataset.stream().filter_nonempty_text_field("code"):
+
+        if "packaging-photo-to-be-selected" in product.get("states", ""):
+            images = get_images(product.get("code"))
```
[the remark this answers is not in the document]
Asteur: Hmm, yes, I know, the first dataset method is encountered to filter products without code (dataset.stream().filter_nonempty_text_field("code")), second get_images() returns you empty result if it won't find necessary data.

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+            if images:
+                images_ids = [
+                    i
+                    for i, j in images.get("product", {}).get("images", {}).items()
```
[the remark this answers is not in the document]
Asteur: Ok

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+                    if not j.get("imgid")
+                ]
```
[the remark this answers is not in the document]
Asteur: I think, it's fine, thanks, I'll save your suggestions for future

robotoff/workers/tasks/update_recycle.py
```diff
@@ -0,0 +1,206 @@
...
+    :return:
+    """
+    url = (
+        "https://world.openfoodfacts.org/api/v0/product/" + ean + ".json?fields=images"
+    )
+    try:
+        r = requests.get(url)
+        if r.ok:
+            return r.json()
+    except Exception:
```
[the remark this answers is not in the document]
Asteur: `requests.RequestException` - ok, good choice
I am not sure, it's necessary to log such actions. 
PS According to other python code and other tasks, it doesn't look like all small steps are logged

robotoff/workers/tasks/update_recycle.py
Asteur: I think it's fine, it is for readability, maybe as you suggested earlier to it in a varibale (`'packaging_{}'.format(lang)`)

robotoff/workers/tasks/update_recycle.py
Asteur: Ok

robotoff/workers/tasks/update_recycle.py
Asteur: It's result of an unselect request

robotoff/workers/tasks/update_recycle.py
Asteur: Thanks, I'll fix it

robotoff/workers/tasks/update_recycle.py
Asteur: I'll see what we can do

robotoff/workers/tasks/update_recycle.py
Asteur: Yes, only if there are suggestions, what to do with them?
Notify user/developer?(we are logging) I don't know robotoff well enough to find other methods to select/unselect images

robotoff/workers/tasks/update_recycle.py
Asteur: Ok, thanks for suggestion

robotoff/workers/tasks/update_recycle.py
Asteur: In general yes, if it helps

robotoff/workers/tasks/update_recycle.py
Asteur: Same as above

alex-marty (approved):

alex-marty:
Awesome, thanks for the new commits! Seems good now!

PR #339, opened 2021-07-23
Fix test_categories.py

kulizhsy:
The failing test is fixed by forcing the cli to read from a particular release, until we have a better story for how the models are released uniformly.

One of the questions I see is why we need this model for a unit test in the first place - ideally you would have a smaller, mock model in order to test the functionality of the predictor rather than using a heavy production model that needs to be fetched from somewhere.

### DIFF
--- a/doc/introduction/contributing.md
+++ b/doc/introduction/contributing.md
@@ -60,10 +60,10 @@ development.
     poetry install
     ```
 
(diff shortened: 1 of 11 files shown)


kulizhsy:
Whoops, too many commits in the wrong branch :) I'm still learning how to use git

PR #341, opened 2021-07-23
Remove hardcoded domains

kulizhsy:
Refactor the codebase to remove hardcoded URLs.

Since we are working on bringing up a 'dev' Robotoff server accessible via 'https://robotoff.openfoodfacts.net', we need to make the URLs used throughout the project more dynamic. To solve this, I am introducing a BaseURLProvider that can construct the desired URL on the fly, accounting for the Robotoff instance the code is currently running on.

### DIFF
--- a/robotoff/cli/batch.py
+++ b/robotoff/cli/batch.py
@@ -48,10 +48,8 @@ def batch_annotate(
             i += 1
             print("Insight %d" % i)
             print(
-                "Add label {} to {}/produit/{}"
-                "".format(
-                    insight.data, BaseURLProvider().country("fr").get(), insight.barcode
-                )
+                "Add label {} to {}/product/{}"
+                "".format(insight.data, BaseURLProvider().get(), insight.barcode)
             )
             print(insight.data)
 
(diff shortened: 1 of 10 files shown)


kulizhsy:
I believe SonerCloud is [misconfigured](https://sonarcloud.io/code?id=robotoff&pullRequest=341) because it's treating `tests/` as a source code path rather than tests, hence the failure

robotoff/cli/batch.py
```diff
@@ -47,8 +48,10 @@ def batch_annotate(
             i += 1
             print("Insight %d" % i)
             print(
-                "Add label {} to https://fr.openfoodfacts.org/produit/{}"
-                "".format(insight.data, insight.barcode)
+                "Add label {} to {}/produit/{}"
```
teolemon, line 51: actually, we need to swap produit as well by product
or do something like fr-en.openfoodfacts.org/product/ ( or even better world.openfoodfacts.org/product/)

Wauplin (approved):

PR #351, opened 2021-08-04
Allow to not post anything on Slack if the robotoff is running as dev/locally

kulizhsy:
This PR achieves the following:

1. SLACK_TOKEN misconfiguration errors occur closer to init time, which helps to catch these misconfigurations sooner.
2. Slack functionality now uses the [factory design pattern](https://realpython.com/factory-method-python/), which allows to better encapsulate slack-only functionality and allows the caller not to care about checking for the currently running instance.
3. Adds tests for the new code (the old code didn't and still does not have test coverage - we can add it incrementally over time).

This is the last change necessary to allow running non-prod versions of Robotoff.

### DIFF
--- a/robotoff/logo_label_type.py
+++ b/robotoff/logo_label_type.py
@@ -1,0 +1,4 @@
+from typing import Optional, Tuple
+
+# Extracted into a separate module to avoid circular dependencies.
+LogoLabelType = Tuple[str, Optional[str]]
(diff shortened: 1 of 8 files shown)


kulizhsy:
There was an unrelated mypy issue from the last PR, the fix for which I added to this PR - hope that's ok

teolemon (approved):
Looks good to me 👍 
Thanks @kulizhsy

kulizhsy:
Thanks!

PR #364, opened 2021-08-17
Remove dead settings

kulizhsy:


teolemon (commented):
Do we need to get rid of the insights.jsonl.gz as well ?

kulizhsy:
Do you mean the actual file? I looked it up on the prod server and it doesn't seem to exist, so it's probably just a setting that wasn't cleaned up properly?

```
kulizhsy@robotoff:/var/lib/docker/volumes/api-dataset$ sudo ls _data
products-etag.txt  products.jsonl.gz  products-min.jsonl.gz
```

teolemon:
So it seems: https://github.com/openfoodfacts/robotoff/commit/c403ecffef3e5d3cdecfb99595d183efc4122c5a

teolemon (approved):

PR #368, opened 2021-08-19
Fix incorrect peewee query

kulizhsy:
Currently many of the API calls will always append 'where latent=false and annotated=false' to the Postgres query because the defaults for these func params are `False` instead of `None`.

This causes missing data to be returned in the API calls. This PR fixes this


kulizhsy:
Also added logging to better understand where the dump API is failing.

teolemon (approved):

teolemon:
Hopefully we're not doing anything silly. CCd @raphael0202 to be sure.

PR #377, opened 2021-08-23
Add code documentation

kulizhsy:
Adding some documentation for things I've discovered while learning how Robotoff operates.

This PR documents:
- The various insight types produced by Robotoff and their current state.
- The ProductInsight table schema.
- The scheduler tasks.
- The worker tasks.


robotoff/insights/_enum.py
```diff
@@ -7,21 +7,75 @@ class StrEnum(str, Enum):
 
 @unique
 class InsightType(StrEnum):
+    """InsightType defines the type of the insight."""
+
+    # The 'ingredient spellcheck' insight corrects the spelling in the given ingredients list.
+    # NOTE: this insight is not currently working - last generated insight was in July 2020.
```
raphael0202, line 13: The import is currently manual for this insight type, which explains why the last generated insight was in July 2020.

robotoff/workers/tasks/__init__.py
```diff
@@ -53,10 +53,38 @@ def delete_product_insights(barcode: str, server_domain: str):
...
+    # When a product is updated, Robotoff will:
+    # 1. Generate new insights related to the product's category and name.
+    # 2. Clean up any unvalidated existing insights that no longer apply to the product.
+    #
     "product_updated": update_insights,
+    # 'object_detection' consists of logo detection: extracting logos from product images and generating logo-related insights.
+    # This task is triggered by the 'import_image' task above.
+    #
     "object_detection": run_object_detection,
+    # 'update_recycling' auto-selects the recycling photos for a given product.
```
raphael0202, line 86: It seems the task is triggered during image import: https://github.com/openfoodfacts/robotoff/blob/48969182ef43799328ea2555082614f5d372021d/robotoff/workers/tasks/import_image.py#L119

robotoff/models.py
```diff
@@ -69,10 +120,16 @@ class ProductInsight(BaseModel):
         "one of 'off', 'obf', 'opff', 'opf'",
         index=True,
     )
+
+    # This field refers to the number of unique IPs that have scanned this product.
     unique_scans_n = peewee.IntegerField(default=0, index=True)
+
+    # A reserved barcode is a barcode that is a 'per-weight' barcode.
+    # NOTE: currently this field is being set for 2*** barcodes, but is not read by anything.
```
raphael0202, line 128: This field is used in the API to filter products with reserved barcodes: https://github.com/openfoodfacts/robotoff/blob/fce2c70308a75c8b8eef406ff5951fdf7735423b/robotoff/app/api.py#L852

robotoff/workers/tasks/__init__.py
```diff
@@ -53,10 +53,38 @@ def delete_product_insights(barcode: str, server_domain: str):
...
 EVENT_MAPPING: Dict[str, Callable] = {
+    # 'import_image' is triggered every time there is a new OCR image available for processing by Robotoff, via /api/v1/images/import.
+    #
+    # On each image import, Robotoff performs the following tasks:
+    #  1. Generates various insights based on the OCR-extracted text from the image.
+    #  2. Extracts the nutriscore insight based on the nutriscore ML model.
+    #  3. Triggers the 'object_detection' task, which is described below.
+    #  4. Stores the imported image metadata in the Robotoff DB.
+    #
+    # NOTE: any insights generated by this call are marked for automatic processing, unlike the insights that are generated
```
raphael0202, line 64: Hum I'm not sure this is true, some insights that are triggered by the webhook call are not automatically processed (such as `product_weight`).

robotoff/workers/tasks/__init__.py
kulizhsy: The link you're referencing is for `object_detection`, but I wasn't not able to find any references for `update_recycling` in the codebase

robotoff/insights/_enum.py
kulizhsy: Thanks, updated the comment!

robotoff/workers/tasks/__init__.py
kulizhsy: Ahh, I was confused by this line:

https://github.com/openfoodfacts/robotoff/blob/837c2da66b80e48a47788dc92d7975bd6c569424/robotoff/workers/tasks/import_image.py#L55


But looking at the code, we only seem to care if this value is `False`:

https://github.com/openfoodfacts/robotoff/blob/837c2da66b80e48a47788dc92d7975bd6c569424/robotoff/insights/importer.py#L230

robotoff/workers/tasks/__init__.py
raphael0202: Sorry my bad, I misread the comment, I thought it was about the object_detection event above

raphael0202: I don't know what is `update_recycling` event though

robotoff/workers/tasks/__init__.py
raphael0202: Well the automatic processing logic is a bit cumbersome, basically there are two import mode: automatic or not: this is specified by the `automatic` field in `import_insights` function. If `automatic` is `False`, no imported insight is automatically applied. This is useful when you batch-import insights from CLI (CLI-import is used for instance when we create a new insight type, to import all insights predicted from the JSONL dump).
If `automatic` is `True`, the insight **may** be automatically applied, depending on the return value of `need_validation` function.

robotoff/workers/tasks/__init__.py
teolemon: it's a new feature that hasn't been deployed to auto-select recycling photos from the existing photos @raphael0202 @kulizhsy 
We intended to test it for a while, and then batch run it to get many recycling images selected, and ready for extraction

robotoff/workers/tasks/__init__.py
kulizhsy: Thanks for the explanation, added a comment to the cli code :)

PR #403, opened 2021-09-01
Use the latest poetry version and invoke gunicorn through it

kulizhsy:
Works locally on my machine - previously using 1.1.8 poetry caused issues on the server, but pinning poetry to 1.1.7 causes missing poetry dependencies at run time.


teolemon (approved):

kulizhsy:
Thanks @teolemon

kulizhsy:
Update: we're not getting any poetry/Docker issues anymore \o/

PR #411, opened 2021-09-01
Add first tests for importer.py and fix automatic processing bug

raphael0202:
- Restore automatic processing for `product_weight`, `expiration_date` insight types
- Add very basic tests for product insight import as a starter

### DIFF
--- a/robotoff/insights/dataclass.py
+++ b/robotoff/insights/dataclass.py
@@ -99,7 +99,7 @@ class Insight:
             data=insight.data,
             value_tag=insight.value_tag,
             value=insight.value,
-            automatic_processing=insight.automatic_processing or False,
+            automatic_processing=insight.automatic_processing,
             barcode=product_insights.barcode,
             source_image=product_insights.source_image,
             predictor=insight.predictor,
--- a/robotoff/insights/importer.py
+++ b/robotoff/insights/importer.py
@@ -143,6 +143,8 @@ class BaseInsightImporter(metaclass=abc.ABCMeta):
         return seen_set
 
     def get_seen_count(self, barcode: str, server_domain: str) -> int:
+        """Return the number of non-latent insights that have the same barcode and
+        server domain as provided as parameter."""
         query = generate_seen_set_query(self.get_type(), barcode, server_domain)
         return query.count()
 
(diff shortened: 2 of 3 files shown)


robotoff/insights/dataclass.py
```diff
@@ -99,7 +99,7 @@ def from_raw_insight(
             data=insight.data,
             value_tag=insight.value_tag,
             value=insight.value,
-            automatic_processing=insight.automatic_processing or False,
+            automatic_processing=insight.automatic_processing,
```
kulizhsy, line 102: [not in scope for this PR, more of a design discussion]: 

It seems to me that the root of the issue is the fact that we're allowing a nullable primitive type when we don't seem to care about the nullability.

I.e. here we're storing this bool as an `Optional[Bool]`, but the database field is not optional and defaults to `False`. 
From a design perspective, it would make sense to have it optional if during the insight processing we cared to differentiate between `True`, `False` and `I don't know`.

Looking at the current usage, this doesn't seem to be the case as we always want to treat `None` as `False`, but because we allow both values it can easily result in bugs like these if you're not super careful.

Does it make sense to try to remove the Optional qualifier from primitives like these (I think 'latent' is in a similar situation) going forward if we're ever refactoring this code?

kulizhsy (approved):
Don't forget to format your changes with black :) (I think @teolemon added an autoblack action but it doesn't seem to be merged yet)

robotoff/insights/dataclass.py
raphael0202: Correct me if I'm wrong, but we do want to differentiate between `True`, `False`, and `I don't know (for now)`. The purpose was to allow insight extractor (OCR-based mostly) to specify the insights for which we have enough confidence to be processed automatically (`True`), the insights for which it is the opposite (False), and to fall-back to default behaviour for this insight type (`None`).

kulizhsy: I see.

I was interpreting it as the insight extractor is the only place where we can make a decision as to whether something is automatically processable since it understands the thresholds for the specific insights. Hence, once the `RawInsight` is constructed nothing else can affect the processability of the insight. But if we want to be able to arbitrarily flip the default then it makes sense I guess.

PR #427, opened 2021-09-10
Clean up the hierarchical classifier training/prediction code

kulizhsy:
Based on discussions with @teolemon and @raphael0202 this code has not been used in quite some time and has been replaced by the neural category classifier in terms of performance. 

Let me know if I should back this file up on `robotoff-ai` or if we're happy to completely remove it


teolemon (approved):
LGTM ✅
Given the poor perf, is it really worth driving people to try to use it ?

PR #433, opened 2021-09-15
docs: Remove old ANN references

ocervell:


.github/workflows/robotoff.yml
```diff
@@ -74,9 +74,6 @@ jobs:
       - name: Code quality isort
         run: poetry run isort . --check
 
-      - name: Typing check mypy
```
kulizhsy, line 77: We still want the mypy check, just without 'services/ann':

`poetry run mypy .`

.github/workflows/robotoff.yml
```diff
@@ -74,9 +74,6 @@ jobs:
       - name: Code quality isort
         run: poetry run isort . --check
```
teolemon, line 76:
```suggestion

  - name: Typing check mypy
  -  run: poetry run mypy . services/ann
```

teolemon: @kulizhsy @ocervell here's a fix

kulizhsy (commented):
Sorry for the confusing initial comment - we want to keep the mypy checks, but just on the '.'

mypy has a [recursive dir checking limitations](https://github.com/python/mypy/issues/8548) so 'services/ann'  was added explicitly to get around it.

PR #435, opened 2021-09-15
Update script to dump the ingredient/category taxonomies used and split to train/test/val subsets.

kulizhsy:


teolemon (approved):
LGTM, with my limited understanding of any potential adverse effect on training. I defer to @raphael0202 as to why we limited languages. FWIW, I believe it's worth trying to train for all languages if not dangerous

kulizhsy:
Just to highlight, the previous version of the script generated the entire dataset **as well as** the per-language sets. This takes quite a bit of time to generate, and we don't really care for the per-language models (the previous model version was [also trained on just the global set](https://github.com/openfoodfacts/openfoodfacts-ai/releases/tag/dataset-category-2020-06-30))/.

I could have also added a arg util to allow specifying the languages on the command line, but removing the them from the default flow was easier.

This PR does not have any negative impacts on the training itself, just makes it easier to construct the dataset.

PR #439, opened 2021-09-16
Fix slack notification typo

kulizhsy:


kulizhsy:
FYI @teolemon  - submitting now to roll out in prod, we're not notifying on automatic processing atm

teolemon:
ok, we disabled notifications for most things, some of them were just at the beginning to test automatic application of labels, for instance.
We enable them when we launch new detections.

PR #444, opened 2021-09-23
Api modifications

alexfauquette:
Because of the translation of the value, the front-end does not have access to the raw value.

So we are not able to query all the questions with the same value.


alexgarel:
@alexfauquette I've made a new PR https://github.com/openfoodfacts/robotoff/pull/579 

Sorry for such a delay.

alexfauquette:
Thanks @alexgarel for the clean PR :)
Don't hesitate to send a message on slack when it's deployed, I will add the associated feature to hungergame

PR #465, opened 2021-10-08
Some refactor in docker

alexgarel:
Here are some, I hope, non controversial docker refactor.

I did not test them well (does the services still works) because I need a hand for that.


alexgarel:
https://github.com/openfoodfacts/robotoff/pull/465#issuecomment-941129620

I don't know what sonnar gates does want from me there ! It is indicating files that I didn't touch…

PR #480, opened 2021-10-27
fix: remove init tasks

ocervell:
depends on #478


alexgarel (approved):
LGTM altough we need #478 to be merged first.

PR #485, opened 2021-11-01
[Feature] Make insight annotation a vote instead of single user decision

zond:
- New AnnotationVote data model.
- Posting answers (when annotating insights) now records username and device ID into a vote instead of annotating the insight.
  - If enough votes are present (3 votes for any option) then a decision is made.
  - If no other annotation has >1 vote, the annotation with 3 votes is saved into the insight, otherwise decision 0 (don't know) is saved.
- Loading questions (when annotating insights) now prioritizes voted-on insights (by sorting them partly by number of votes) to reduce time to get partly annotated insights finished.
- Loading questions (when annotating insights) now filters out insights the user or device have voted on previously.
- New test target for integration tests requiring the entire dev environment to test the endpoint changes.


.env
```diff
@@ -12,7 +12,7 @@ OFF_GID=1000
 # Robotoff
 TAG=latest
 ROBOTOFF_INSTANCE=localhost
-ROBOTOFF_DOMAIN=openfoodfacts.localhost
+ROBOTOFF_DOMAIN=openfoodfacts.net
```
zond, line 15: This is _only_ used (AFAIU) to download a JSON file from PO.

I'm not very happy about it - what alternatives do we have?

Makefile
```diff
@@ -72,6 +72,7 @@ log:
 
 dl-models:
 	@echo "🥫 Downloading models …"
+	chmod a+w models
```
zond, line 75: I added this just to get my `make` command when first checking out the code to work. Has anyone else experienced that problem? This doesn't feel like the correct solution.

docker/dev.yml
```diff
@@ -23,4 +23,7 @@ x-robotoff-dev: &robotoff-dev
 services:
   api: *robotoff-dev
   workers: *robotoff-dev
-  scheduler: *robotoff-dev
\ No newline at end of file
+  scheduler: *robotoff-dev
+  postgres:
+    ports:
+      - 15432:5432
```
zond, line 29: We might want to remove this, or does anyone think it might be useful in the future?

zond:
It annoys me a lot that we have GitHub actions that fail due to things outside this PR. How prioritized is it to get those things working?

.env
kulizhsy: Not only for JSON - look at the usage of `BaseURLProvider`.

What are you not happy about in particular, the file downloads/the fact we have to manipulate URLs to get things working?

docker/dev.yml
kulizhsy: Since we can access the db through docker, I'm not sure if we also want to expose the ports

.env
zond: Hm, maybe I should have clarified - I meant in the integration tests. But it might not be true even there of course :)

What I'm unhappy with is how the test isn't hermetic - if the .net environment is down the tests will fail.

docker/dev.yml
zond: Alright, I'll remove it.

.env
kulizhsy: We can construct a test taxonomy JSON with the one category we need in tests and mock out the actual call, returning the test taxonomy instead?

kulizhsy: Actually, seems we can just use a [similar set up to this test](https://github.com/openfoodfacts/robotoff/blob/837c2da66b80e48a47788dc92d7975bd6c569424/tests/test_taxonomy.py#L9), since Robotoff already has a fallback taxonomy JSON stored

zond: That's a promising idea, but the test you linked doesn't make the API in general use a different source for JSON files, it just validates that using those local files works, right?

I'm looking at how to tell the entire Robotoff engine to use the local files instead..

zond: That's a promising idea, but the test you linked doesn't make the API in general use a different source for JSON files, it just validates that using those local files works, right?

I'm looking at how to tell the entire Robotoff engine to use the local files instead..

zond: Alright, when I started looking (and trying to make the taxonomy JSON specifically load locally) I noticed that Robotoff uses PO to load product images and things as well.

I think it will have to rely on the .net environment until we get enough problems from it that it's worth either mocking up a productopener, or forcing the integration tests to run a local productopener as well.

docker/dev.yml
ocervell: +1

zond:
I tried to do it the standard git way, using `git rebase master`, but it lead to horrible and repeated merge problems that just didn't happen when I tried a simple `git merge master` instead. I'm OK with `merge` instead of `rebase` (I prefer it even), but anyone complaining can please explain to me what I did wrong first :)

robotoff/app/core.py
```diff
@@ -78,24 +99,34 @@ def get_insights(
 
     if order_by is not None:
         if order_by == "random":
-            query = query.order_by(peewee.fn.Random())
+            query = query.order_by((peewee.fn.Random() * ProductInsight.n_votes).desc())
```
alexgarel, line 102: I'm curious if this is effective enough on a large number of insight, to make voted one really get first places ? (I can see it multiplies expected value, but I have the feeling that on a large collection, we won't get so much voted results !)

robotoff/app/core.py
```diff
@@ -78,24 +99,34 @@ def get_insights(
...
 
-def save_insight(
+def save_annotation(
     insight_id: str,
     annotation: int,
+    device_id: str,
     update: bool = True,
     data: Optional[Dict] = None,
     auth: Optional[OFFAuthentication] = None,
+    verify_annotation: bool = False,
```
alexgarel, line 123: Opinion: I find the name a bit misleading. maybe verified_annotation (True meaning we do not use voting) could be more explicit. Or trusted_annontation.

robotoff/app/core.py
```diff
@@ -107,5 +138,56 @@ def save_insight(
...
+        #  2) 1st place and 2nd place have 2 votes each.
+        #
+        # In both cases, we consider this an ambiguous result and mark it with 'I don't know'.
+        if (
+            existing_votes[0].num_votes >= 2
+            and len(existing_votes) > 1
+            and existing_votes[1].num_votes >= 2
+        ):
+            # This code credits the last person to contribute a vote with a potentially not their annotation.
+            annotation = 0
```
alexgarel, line 186: we do not seem to have an enum for this, a pity !

robotoff/app/core.py
```diff
@@ -107,5 +138,56 @@ def save_insight(
...
+        verified: bool = False
+
+        AnnotationVote.create(
+            insight_id=insight_id,
+            username=auth.get_username() if auth else None,
+            value=annotation,
+            device_id=device_id,
+        )
+
+        with db.atomic() as tx:
```
alexgarel, line 151: why don't we run all the save_annotation function atomically ? (that would make more sense).

alexgarel: in fact in annotator.annotate we will open a distinct transaction. But doesn't that mean, we can be in an inconsistent state ?

robotoff/app/core.py
```diff
@@ -107,5 +138,56 @@ def save_insight(
     if insight.annotation is not None:
         return ALREADY_ANNOTATED_RESULT
 
+    if verify_annotation:
+        verified: bool = False
```
alexgarel, line 142: Opinion: that would be better to have a separate method for this part.

robotoff/app/middleware.py
```diff
@@ -3,7 +3,7 @@
 
 class DBConnectionMiddleware:
     def process_resource(self, req, resp, resource, params):
-        db.connect()
+        db.connect(reuse_if_open=True)
```
alexgarel, line 6: Nice.

alexgarel (commented):
I didn't review the tests yet, but prefer to send my comments so far.

I'm rather picky in my comments, so feel free to ignore some of them.

Really cool to have distinguished integration tests from unit tests.

robotoff/models.py
```diff
@@ -74,7 +76,11 @@ class ProductInsight(BaseModel):
     # 1 = Validated
     annotation = peewee.IntegerField(null=True, index=True)
 
-    # If the insight was annotated manually, this field stores the username of the annotator.
+    # The number of votes for this annotation.
+    n_votes = peewee.IntegerField(null=False, index=True)
```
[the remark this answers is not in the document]
alexgarel: I've seen why in the code (we wan't to sort on it), so documenting is enough.

Makefile
zond: Should I have done that manually? Is there no way to do it automatically?

robotoff/models.py
```diff
@@ -146,6 +152,22 @@ def create_from_latent(cls, latent_insight: "ProductInsight", **kwargs):
         return cls.create(**updated_values)
 
 
+class AnnotationVote(BaseModel):
+    id = peewee.UUIDField(primary_key=True, default=uuid.uuid4)
```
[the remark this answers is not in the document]
zond: I tried to not even provide `id`, but it didn't work - and since I'm more used to autoincrementing ints for PKs, I thought this was the way you do it with UUIDs. What is the correct way?

robotoff/models.py
```diff
@@ -146,6 +152,22 @@ def create_from_latent(cls, latent_insight: "ProductInsight", **kwargs):
...
+    # The insight this vote belongs to.
+    insight_id = peewee.ForeignKeyField(ProductInsight, null=False, backref="votes")
+    # The username of the voter - if logged in.
+    # Currently logged in users do not need to vote on an insight for their annotaion to
+    # be applied.
+    username = peewee.TextField(index=True, null=True)
+    # The value of the annotation, see ProductInsight.annotation.
+    value = peewee.IntegerField(null=False)
```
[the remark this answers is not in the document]
zond: Added `choices`.

robotoff/models.py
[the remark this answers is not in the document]
zond: Added documenting comment.

robotoff/app/core.py
```diff
@@ -67,6 +71,23 @@ def get_insights(
 
     query = ProductInsight.select()
 
+    def join_with_no_votes_by(query, criteria):
+        return query.join(
+            AnnotationVote,
+            join_type=peewee.JOIN.LEFT_OUTER,
+            on=((AnnotationVote.insight_id == ProductInsight.id) & (criteria)),
+        ).where(AnnotationVote.id.is_null())
```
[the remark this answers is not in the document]
zond: Mostly because it's only used inside this particular function. If you feel strongly about I'll change it, but it works with my taste ;)

robotoff/app/core.py
```diff
@@ -67,6 +71,23 @@ def get_insights(
...
+
+    if avoid_voted_by_username:
+        query = join_with_no_votes_by(
+            query, (AnnotationVote.username == avoid_voted_by_username)
+        )
+
+    if avoid_voted_by_device_id:
+        query = join_with_no_votes_by(
+            query, (AnnotationVote.device_id == avoid_voted_by_device_id)
+        )
```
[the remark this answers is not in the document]
zond: It would actually work with both, but it would be a potentially more expensive JOIN. We could of course enforce that it's not used that way, but it didn't feel important. WDYT?

robotoff/app/core.py
zond: If I'm not mistaken, insights with n+1 votes will be (n+1)/(n) more likely to bubble to the top than insights with n votes. It's not perfect, but it doesn't need to be. I suggest we wait and see how it ends up working?

robotoff/app/core.py
zond: I also found it a bit confusing, but was happy with the comment in the beginning of `save_annotation`. I defer to @kulizhsy for this discussion.

robotoff/app/core.py
zond: Is this you wanting us to change something in the PR, or is it just a sentiment about the depressing state of the world? :)

robotoff/app/core.py
zond: The code would be more complex if we wanted all of it in a transaction, and transactions are expensive.

At the same time, the only thing important to keep atomic there is that the n_votes is correct, so that insights don't stay at the bottom of the pile by mistake.

If two votes arrive at the same time, the worst thing that could happen currently is that the second vote would override the first vote, but I'm not sure that is bad enough to warrant extra code complexity.

WDYT?

robotoff/app/core.py
zond: I think that's a part of a reasonable cleanup with the transactions as well. A proper refactoring would probably split it out into one function that made votes and optionally also called a function that annotated insights if enough votes were collected - and the second function would either reuse or create a new transaction as needed.

Do we need it for this PR though?

robotoff/app/middleware.py
zond: Yeah, it was necessary to be able to run the integration tests. Not sure why - but it seems the app starts up twice when I run them..?

robotoff/models.py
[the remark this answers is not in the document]
kulizhsy: I'm not sure if it's a peewee issue or a postgres issue, but there indeed does not seem to be a way to just leave it unset, without a default, and have an ID auto-assigned.

I saw an alternative way to have it auto-assigned with   `peewee.UUIDField(.. ,constraints=[SQL("DEFAULT uuid_generate_v4()")])`

but that also was throwing errors for me..

robotoff/app/core.py
kulizhsy: I like Alex's suggestion of the name `verified_annotation` & flipping of the default.

zond: @kulizhsy will you change it or should I?

robotoff/models.py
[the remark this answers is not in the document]
zond: Let's leave it like this then. Feel free to reopen the conversation if you disagree :D

robotoff/app/core.py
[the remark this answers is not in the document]
zond: @kulizhsy does that mean you volunteer to implement it? :)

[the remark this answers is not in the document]
kulizhsy: Sure, I'll do it :)

robotoff/app/core.py
kulizhsy: Sure, let's go with `trusted_annotator`, I'll update the code.

robotoff/app/core.py
alexgarel: yep, it's just that if you have 1000 insight, and 10 with 2 votes, you will only have  20 / 1010 chances of getting an already voted one. But You're right that this has the advantage of being very simple :-)
Otherwise I was thinking something like:
random / ((4 - votes) * alpha)
where alpha could be 4

robotoff/app/core.py
alexgarel: I'm ok with it.
As you say there no really evil consequences.

.github/workflows/pull_request.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       # Launch tests
       #----------------------------------------------
       - name: Launch tests
-        run: poetry run pytest --cov-report xml --cov=robotoff tests
+        run: make tests
```
kulizhsy, line 104: This seems to break the github action:

`Service 'workers' failed to build : failed to export image: failed to create image: failed to get layer sha256:b5c7da884f67639fd3dec4ab67e5950371fdfed950da68b3dc0b40f5199d8682: layer does not exist`

[Full log](https://github.com/openfoodfacts/robotoff/runs/4093308717?check_suite_focus=true)

@alexgarel @zond how do we fix this?

robotoff/app/core.py
alexgarel: It was a gentle push about if you feel like fixing it in this PR (if you were already on the edge of doing it).

.github/workflows/pull_request.yml
zond: Can it be temporary network problems? I looks like the docker image repo was broken.

kulizhsy: Not sure, but it was[ failing yesterday](https://github.com/openfoodfacts/robotoff/runs/4066110863?check_suite_focus=true) too

zond: Does it work locally for you? For me they do.

@alexgarel, @ocervell do you know why the tests would fail due to failure to get a docker image - on GitHub but not locally...?

tests/integration/test_api.py
```diff
@@ -0,0 +1,336 @@
...
+    votes = list(AnnotationVote.select())
+    assert len(votes) == 0
+
+    insight = next(
+        ProductInsight.select()
+        .where(ProductInsight.id == insight_id)
+        .dicts()
+        .iterator()
+    )
+    assert insight.items() > {"username": "a", "annotation": -1, "n_votes": 0}.items()
```
alexgarel, line 147: wow, never seen that kind of dict_items comparison before ! That's cool.

tests/integration/test_api.py
```diff
@@ -0,0 +1,336 @@
...
+
+    insight = next(
+        ProductInsight.select()
+        .where(ProductInsight.id == insight_id)
+        .dicts()
+        .iterator()
+    )
+
+    assert all(
+        [
```
alexgarel, line 184: you don't need to put list "[" in all, you can directly write:
`all(val is None for key, val in insight.items() if key in {"username", "completed_at", "annotation"})`

I would even have written:
`not any(insight[key] for key in ("username", "completed_at", "annotation"))`

tests/integration/test_api.py
```diff
@@ -0,0 +1,336 @@
...
+
+    insight = next(
+        ProductInsight.select()
+        .where(ProductInsight.id == insight_id)
+        .dicts()
+        .iterator()
+    )
+    # The insight should be annoted with '0', with a None username since this was resolved with an
+    # anonymous vote.
+    assert insight.items() > {"annotation": 0, "username": None, "n_votes": 5}.items()
```
alexgarel, line 336: Good test covarage ! :+1:

alexgarel (approved):
Ok, some more pedantic comments, but overall it's ok !

robotoff/app/api.py
```diff
@@ -805,15 +819,29 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, barcode: str):
         response: JSONType = {}
         count: int = req.get_param_as_int("count", min_value=1) or 1
         lang: str = req.get_param("lang", default="en")
+        # If the device_id is not provided as a request parameter, we use the
+        # hash of the IPs as a backup.
+        device_id: str = req.get_param(
+            "device_id",
+            default=hashlib.sha1(str(req.access_route).encode()).hexdigest(),
+        )
```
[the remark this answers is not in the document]
zond: Fixed it.

Makefile
zond: Super annoying, there should always be a default that just works I feel. But I'll remove the chmod :)

tests/integration/test_api.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+def test_barcode_question():
+    cl = client()
+    result = cl.simulate_get("/api/v1/questions/2")
+
+    assert result.status_code == 200
+    assert result.json == {"questions": [], "status": "no_questions"}
+
+    result = cl.simulate_get("/api/v1/questions/1")
```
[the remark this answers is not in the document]
zond: Deferring this to @kulizhsy

tests/integration/test_api.py
zond: What does it actually do? I sort of feel that it should check that it has "at least those elements", but I can't say I feel certain about it.

kulizhsy: @zond, yep, it does what you said :)

zond: Nice one, thanks!

alexgarel: Yes python docs says dict_items acts as a Set of (key, value). And for sets, A > B means that A got all the elements from B and some more.

Makefile
ocervell: I don't think we ***need*** to set `OFF_UID` for Robotoff --> this adds a step that is slowing local dev imo. For PO we do need to change ownership of data folders coming from external, but for Robotoff I don't think it should be necessary.

.github/workflows/pull_request.yml
alexgarel: Seems like a known bug see https://stackoverflow.com/questions/51115856/docker-failed-to-export-image-failed-to-create-image-failed-to-get-layer#62409523 and https://github.com/moby/moby/issues/37965
Some suggest adding a `RUN true` to Dockerfile after `COPY` or using BUILDKIT (which I would like :-) )

@ocervell what do you think ?

ocervell: Never seen this issue before, we can try using BUILDKIT by setting the env variable `DOCKER_BUILDKIT=1` in the step `env`.

Makefile
alexgarel: I disagree, in the sense, that you will end up with folders on your machine, that you can't list / copy etc. without using sudo, which seems worst to me. (but yes it's a compromise to be found)

For me as soon as you:
- don't use root in your container (security best practice)
- are a local dev (not on machine, where you handle multi user and usually connect with root / sudoer) (this may apply if you admin stuff with non sudoers users on the system)
- bind mount folders in docker

You have to align container user and your id.

Typically here, for the models folders, how do you handle it otherwise, appart from setting `chmod a+x models/` or `sudo chown 1000:1000 models/` ?

tests/integration/test_api.py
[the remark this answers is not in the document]
kulizhsy: I'll update the tests

tests/integration/test_api.py
kulizhsy: Done, thanks!

.github/workflows/pull_request.yml
kulizhsy: Hmm, `DOCKER_BUILDKIT=1` didn't seem to help :)

Makefile
zond: Git only seems to care about executable permissions unfortunately.

Do we even _need_ OFF_UID for Robotoff? Why not just run everything as current user?

robotoff/app/core.py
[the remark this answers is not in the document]
zond: No more comments received, so I guess I'll resolve this.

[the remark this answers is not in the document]
zond: Ah, I see it's also redundant after another change :D

robotoff/app/core.py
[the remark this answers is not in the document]
zond: And I see it's done. Nice!

zond:
We only have very few unresolved comments now, almost ready to merge to master. However, the automatic testing still doesn't work. In particular `DeepSource` and `Robotoff code quality checks and unit tests`. Does anyone have ideas why?

stephanegigandet (approved):

.github/workflows/pull_request.yml
alexgarel: I added the `RUN true`, let see if it fails or not…

alexgarel:
@kulizhsy  
1. I lowered requirements in deepsource on documentation, and tell it not to block PRs (let see if it works)
2. I fixed the problem with docker
3. I changed the way integration tests are runned (exec is not the good solution from what I know), and removed the `-v` to `docker down` as I want to run tests locally without removing my volumes !

Now if I run tests locally, I run into problems because of pewee having a failed transaction. I think it should be handled by the pytest fixture : robustly rollback any pending transaction after the script.

robotoff/models.py
```diff
@@ -74,7 +76,12 @@ class ProductInsight(BaseModel):
     # 1 = Validated
     annotation = peewee.IntegerField(null=True, index=True)
 
-    # If the insight was annotated manually, this field stores the username of the annotator.
+    # The number of votes for this annotation.
+    # Stored here for quick sorting.
+    n_votes = peewee.IntegerField(null=False, index=True)
+
```
alexgarel, line 82: (sorry I comment here to have a new thread).

We add a column, but how do we handle database migration ?

Integration tests are failing in my case because my database already exists and does not have this column. Did I missed some step ?

Maybe we should use http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#schema-migrations

Makefile
```diff
@@ -108,10 +108,18 @@ checks: flake8 black-check mypy isort-check docs
...
 	@echo "🥫 Running tests …"
 	# run tests in worker to have more memory
-	${DOCKER_COMPOSE} run --rm workers poetry run pytest tests
+	${DOCKER_COMPOSE} run --rm workers poetry run pytest --cov-report xml --cov=robotoff tests/unit 
+
+integration-tests: dev
+	@echo "🥫 Running integration tests …"
+	# run tests in worker to have more memory
+	${DOCKER_COMPOSE} run --rm workers poetry run pytest -vv --cov-report xml --cov=robotoff --cov-append tests/integration
+	${DOCKER_COMPOSE} down
```
alexgarel, line 122: The way we run integrations tests, without changing database name, means we are destructive on the dev desktop. 

I think we can tackle that as a separate issue, but do you agree this is a problem ? (I do find it an important problem)

.github/workflows/pull_request.yml
alexgarel: This does the trick, now let see if it pass or fail for other reasons.

robotoff/models.py
alexgarel: @zond **this is a blocker**

zond: As I said in the chat - I think we could migrate manually this time. AFAIU this is the first ever DB change in Robotoff, and if migrations only happen every 3 years maybe we don't need to build a migration system this time?

An alternative is to build migration support in a separate PR, and merge that one into this, and add a proper migration.

WDYT?

Makefile
alexgarel: I do :-)

robotoff/models.py
zond: IIRC we agreed to do this migration manually? Closing this discussion.

Makefile
zond: Alright, let's leave that as topic for a new PR?

alexgarel: ok.

alexgarel: I opened an issue : https://github.com/openfoodfacts/robotoff/issues/490

robotoff/models.py
```diff
@@ -74,7 +76,12 @@ class ProductInsight(BaseModel):
     # 1 = Validated
     annotation = peewee.IntegerField(null=True, index=True)
 
-    # If the insight was annotated manually, this field stores the username of the annotator.
+    # The number of votes for this annotation.
+    # Stored here for quick sorting.
```
alexgarel, line 80: @zond @kulizhsy here is the SQL I'am planning to run to migrate:

```
ALTER TABLE product_insight ADD COLUMN IF NOT EXISTS n_votes integer NOT NULL DEFAULT 0;
```
I think the best is to run it ahead of deployment, as it should not disturb.

I am wondering, if I should also run a :
```
UPDATE product_insight SET n_votes=1 where annotation = 1;
```

zond: The migration SQL seems correct.

I don't think you need to run the other code - we'll just have to keep in mind that old annotations are annotated without votes. Setting `n_votes` without actually populating the vote table seems dangerous. Better to keep `n_votes` at zero and use it to identify old annotations without votes?

ocervell:
@alexgarel can we merge this ? Are there opened issues still ?

robotoff/models.py
alexgarel: I've run the migration on both preprod and actual prod (10.1.0.103)

cc: @ocervell

alexgarel:
@ocervell @zond and @kulizhsy 

I would be ready to Merge, now it seems to me, it would be better to have some time to test it in preprod.

And as we are about to deploy to new prod (probably on monday), I wonder if we should wait to merge this one after that ?

PR #496, opened 2021-11-19
Fix logo annotation Slack alerting

kulizhsy:
Purposefully did not add tests to the import_image code as I don't understand what it should be doing.

Hopefully this is OK for the time being, we should probably take a better look at ANN anyway, since it seems to be having some issues.


alexgarel (approved):
LGTM (only a quick check).

Issue #412, opened 2021-09-02
Compress photos when sent to slack


CharlesNepote:
When Robotoff detects an image which might have to be moderated (faces, etc.), it sends it to a private Slack chanel to alert moderators. Some images are too big for Slack and it doesn't show it automatically, it has to be expanded. See the second item in this example:
![image](https://user-images.githubusercontent.com/562352/131804906-391d13bd-3185-46f3-b129-de3138316330.png)

Expand these photos takes some time. It's very annoying in the moderation process. Robotoff should send compressed photos to Slack, to be sure there are automatically shown.

CharlesNepote:
I'm not sure, but I think every image has already a medium and a small version:
https://images.openfoodfacts.org/images/products/762/221/044/9283/1.jpg (original image)
https://images.openfoodfacts.org/images/products/762/221/044/9283/1.400.jpg (medium)
https://images.openfoodfacts.org/images/products/762/221/044/9283/1.100.jpg (small)

You just have to add ".400" after the number of the photo.

A structured block would be also very useful. I played with your example and [tried to enhance it](https://app.slack.com/block-kit-builder/T02KVRT1Q#%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22type:%20label_annotation%20%5Cnlabel:%20*thumb*,%20score:%200.6113166%5Cn%20%3Chttps://static.openfoodfacts.org/images/products/762/220/142/5401/1.jpg%7CImage%3E%20--%20%3Chttps://world.openfoodfacts.org/cgi/product.pl?type=edit&code=8436546053682%7C*Edit*%3E%22%7D,%22accessory%22:%7B%22type%22:%22image%22,%22image_url%22:%22https://static.openfoodfacts.org/images/products/762/220/142/5401/1.400.jpg%22,%22alt_text%22:%22-%22%7D%7D%5D%7D).

PR #497, opened 2021-11-19
Add structured messages to Slack alerting

kulizhsy:
An example of the new image alerting can be found [here](https://app.slack.com/client/T02KVRT1Q/C02P0PR5VS4)

### DIFF
--- a/robotoff/slack.py
+++ b/robotoff/slack.py
@@ -50,6 +50,56 @@ class NoopSlackNotifier(SlackNotifierInterface):
     pass
 
 
+def _sensitive_image(flag_type: str, flagged_label: str) -> bool:
+    """Determines whether the given flagged image should be considered as sensitive."""
+    return (
+        flag_type == "label_annotation"
+        and flagged_label
+        in {
+            "face",
+            "head",
+            "selfie",
+            "hair",
+            "forehead",
+            "chin",
+            "cheek",
+            "tooth",
+            "eyebrow",
+            "ear",
+            "neck",
+            "jaw",
+            "nose",
+            "facial expression",
+            "glasses",
+            "eyewear",
+            "child",
+            "baby",
+            "human",
+        }
+    ) or flag_type == "safe_search_annotation"
+
+
+def _slack_message_block(
+    message_text: str, with_image: Optional[str] = None
+) -> List[Dict]:
+    """Formats given parameters into a Slack message block."""
+    block = {
+        "type": "section",
+        "text": {
+            "type": "mrkdwn",
+            "text": message_text,
+        },
+    }
+
+    if with_image:
+        block["accessory"] = {
+            "type": "image",
+            "image_url": {with_image},
+            "alt_text": "-",
+        }
+    return [block]
+
+
 class SlackNotifier(SlackNotifierInterface):
     """SlackNotifier implements the real SlackNotifier."""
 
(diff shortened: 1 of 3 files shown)


robotoff/slack.py
```diff
@@ -50,6 +51,56 @@ class NoopSlackNotifier(SlackNotifierInterface):
...
+            "jaw",
+            "nose",
+            "facial expression",
+            "glasses",
+            "eyewear",
+            "child",
+            "baby",
+            "human",
+        }
+    ) or flag_type == "safe_search_annotation"
```
alexgarel, line 80: to ease reading:

```suggestion
    is_human : bool = flagged_label in {
        "face",
        "head",
        "selfie",
        "hair",
        "forehead",
        "chin",
        "cheek",
        "tooth",
        "eyebrow",
        "ear",
        "neck",
        "jaw",
        "nose",
        "facial expression",
        "glasses",
        "eyewear",
        "child",
        "baby",
        "human",
    }
    return (is_human and flag_type == "label_annotation") or flag_type == "safe_search_annotation"
```

robotoff/slack.py
```diff
@@ -72,140 +123,72 @@ class SlackNotifier(SlackNotifierInterface):
...
+                text += f"type: {flag_type}\nlabel: *{label}*, match: {match_text}\n"
 
-        url = settings.OFF_IMAGE_BASE_URL + source
-        edit_url = "{}/cgi/product.pl?type=edit&code={}" "".format(
-            settings.BaseURLProvider().get(), barcode
-        )
-        text += url + "\n"
-        text += "edit: {}".format(edit_url)
+        edit_url = f"{settings.BaseURLProvider().get()}/cgi/product.pl?type=edit&code={barcode}"
+        image_url = settings.OFF_IMAGE_BASE_URL + source_image
```
alexgarel, line 157: It seems we could squash https://github.com/openfoodfacts/robotoff/issues/412 by just adding a transformation here:
above:

```python
define low_res_image(source_image, version="400"):
    return source_image.replace(".jpg", "." + version + ".jpg")
```
here:

```python
        image_url = settings.OFF_IMAGE_BASE_URL + low_res_image(source_image)
```

The 400px version is auto generated by the server for every images.

alexgarel: (you would have to also add a .jpg to your source_image in your tests)

tests/test_slack.py
```diff
@@ -17,3 +56,107 @@ def test_slack_token(t: str) -> str:
     monkeypatch.setattr(settings, "slack_token", lambda: test_slack_token(token_value))
     notifier = slack.NotifierFactory.get_notifier()
     assert type(notifier) is want_type
+
+
```
alexgarel, line 60: Could be good to have a test where we send nothing !

robotoff/slack.py
```diff
@@ -234,26 +217,22 @@ def send_logo_notification(
             f"(<https://hunger.openfoodfacts.org/logos?logo_id={logo.id}|annotate logo>, "
             f"<{base_off_url}/product/{barcode}|product>):\n{prob_text}"
         )
-        self._post_message(text, self.ROBOTOFF_ALERT_CHANNEL)
+        self._post_message(_slack_message_block(text), self.ROBOTOFF_ALERT_CHANNEL)
 
     def _post_message(
```
alexgarel, line 222: Off topic:
I think it would be better if the NoopSlackNotifier would inherit the SlackNotifier but modify _post_message to just log the payload as "info".
It would help spot eventual bugs in SlackNotifier earlier.

alexgarel (approved):
LGTM with some suggestions

robotoff/slack.py
kulizhsy: With the migration to the blocks API, we don't need to manually downscale the image - see[ the last message in the test channel](https://app.slack.com/client/T02KVRT1Q/C02P0PR5VS4).

For the image link, we probably want the full-scale image anyway. If the alert image looks suspicious from the Slack channel clicking on image link should add more detail to the image rather than show the same resolution as in the thumbnail.

tests/test_slack.py
kulizhsy: Added!

robotoff/slack.py
kulizhsy: Good point, I'll send a separate PR for that to keep this PR focused!

PR #498, opened 2021-11-23
Noop Slack Notifier inherits from the real Notifier

kulizhsy:


alexgarel (approved):
Really cool.

robotoff/slack.py
```diff
@@ -244,6 +238,24 @@ def _post_message(
             )
 
 
+class NoopSlackNotifier(SlackNotifier):
+    """NoopSlackNotifier is a NOOP SlackNotifier used in dev/local executions of Robotoff."""
+
+    def __init__(self):
+        pass
```
alexgarel, line 245: To avoid deepsource error, maybe use:

```suggestion
    def __init__(self):
        super().__init__("")
```

(but really, I think deepsource is really tedious here)

kulizhsy:
Thanks @alexgarel!

PR #500, opened 2021-11-24
Disable automatic processing of Nutri-Score grade insights until we h…

kulizhsy:
…ave more confidence in the output & add tests


kulizhsy:
Ideally we would control automatic processing/thresholds via env vars/a config such that we can quickly disable Robotoff when things start to go badly.

But this would require significant code changes, so I went with the simpler option for now

alexgarel:
@kulizhsy : I have lowered DeepSource settings, because it is really too picky ! I rerun it, you should be able to merge after that.

kulizhsy:
@alexgarel nice, thanks!

PR #505, opened 2021-11-29
Add the default doc type to ES requests

kulizhsy:
Turns out you still need to specify the doctype in ES [insert requests](https://elasticsearch-py.readthedocs.io/en/v7.15.2/api.html#elasticsearch.Elasticsearch.bulk) even if it's the default type.


kulizhsy:
Merging now to unblock dev Robotoff set-up verification. @alexgarel let me know if you have any concerns!

PR #507, opened 2021-11-30
Re-introduce DOCTYPE for Elasticsearch queries

kulizhsy:
In #478 I removed the doc_type parameters as they were supposedly not needed in later ES versions.

However, even with the default `_doc`, we have to provide the document type when making requests.
Since we need to provide it anyway, I've reverted `_doc` to `document` as keeping it as `_doc` would require a prod migration. As we're trying to migrate prod to a new set-up, any additional complexities are better avoided at this point


kulizhsy:
Let me know if you have any concerns @alexgarel, merging to unblock the prod Robotoff rollout

PR #512, opened 2021-12-02
feat: use the new category classifier model for category prediction

kulizhsy:


robotoff/cli/main.py
```diff
@@ -194,25 +194,31 @@ def download_models(force: bool = False) -> None:
...
     if predicted:
-        if blacklist:
-            predicted = filter_blacklisted_categories(predicted)
-
-        for cat, confidence in predicted:
-            print("{}: {}".format(cat, confidence))
+        for prediction in predicted:
+            print(f"{prediction.category}: {prediction.confidence}")
+    else:
+        print("Nothing predicted")
```
alexgarel, line 221: shouldn't we add the barcode in the printed messages ?

robotoff/taxonomy.py
```diff
@@ -136,16 +145,31 @@ def iter_nodes(self) -> Iterable[TaxonomyNode]:
...
-                    excluded.add(second_node.id)
+        if not ignore_ancestor_gaps:
+            excluded = _ancestor_set(nodes)
+        else:
+            for node in nodes:
+                for second_node in (
+                    n for n in nodes if n.id not in excluded and n.id != node.id
+                ):
+                    if node.is_child_of(second_node):
+                        excluded.add(second_node.id)
```
alexgarel, line 172: Is this working ? I don't think this will take into account ancestor_gaps, for is_child_of is recursive and not limited to those nodes.

Having unit tests would be useful

alexgarel (approved):
@kulizhsy sorry I've only done a quick read.

However I validate because I won't be there until wednesday.

Still I think you have to fix find_deepest_nodes.

Also I would like you to confirm: most of the code you removed here is now on the training side in robotoff-ann, in the form of tensorflow preprocessing steps (or do I misunderstand ?)

robotoff/cli/main.py
kulizhsy: Done

robotoff/taxonomy.py
kulizhsy: Sorry, I somehow didn't push all of the changes in this PR - you're correct, since `is_child_of` is recursive we don't need the additional flag. Added the missing changes to the PR.

kulizhsy:
@alexgarel,

> Also I would like you to confirm: most of the code you removed here is now on the training side in robotoff-ann, in the form of tensorflow preprocessing steps (or do I misunderstand ?)

`robotoff-ann` is unrelated to TF models - it is a service that learns/predicts logos from images, nothing else. `robotoff-ann`  is not using TensorFlow, it uses EfficientNet, which I believe is PyTorch.

I didn't move code to the *training* side - the functionality of the deleted code is now being carried out by the TF model via [Keras pre-processing layers](https://github.com/kulizhsy/off-category-classification/blob/master/category_classification/models.py#L78), i.e. the *serving* side.

alexgarel:
@kulizhsy ok for me. Ok also for the tf models, I mess things a bit in my understanding.

PR #517, opened 2021-12-08
Replace moving of the PO dump from /tmp with copy

kulizhsy:
Based on the slack discussion:

`shutil.move` throws a `Invalid cross-device link:` when moving files from /tmp/ to /opt/robotoff/.. This is most likely a result of shutil.move optimisations.

Instead of moving the file in one step, we can bypass the issue with an explicit `copy` followed by a `remove` operation.
In this case, an explicit `remove` is not necessary as cleanup will happen when we exist the `with ....` branch.


ocervell (approved):
Sorry, I read too fast :)

PR #518, opened 2021-12-09
fix: permissions on datasets folder

ocervell:


Dockerfile
```diff
@@ -46,6 +46,7 @@ RUN cd /opt/robotoff/i18n && \
     bash compile.sh
 COPY --chown=off:off robotoff /opt/robotoff/robotoff/
 COPY --chown=off:off data /opt/robotoff/data
+COPY --chown=off:off datasets /opt/robotoff/datasets
```
alexgarel, line 49: Hum, I don't think we want to copy datasets in the image.

I think there it is better to have a :
```
RUN mkdir -p /opt/robotoff/datasets && chown off:off -R /opt/robotoff/datasets
```

PR #527, opened 2021-12-13
feat: add health tests & improve CI

ocervell:


.env
```diff
@@ -24,14 +24,14 @@ POSTGRES_USER=postgres
 POSTGRES_PASSWORD=postgres
 
 # InfluxDB
-INFLUXDB_HOST=localhost
+INFLUXDB_HOST=host.docker.internal
```
alexgarel, line 27: `host.docker.internal` one only works on mac or windows, not on linux

See also mongo uri down there.

.github/workflows/container-deploy.yml
```diff
@@ -131,6 +131,24 @@ jobs:
...
+        username: ${{ secrets.USERNAME }}
+        key: ${{ secrets.SSH_PRIVATE_KEY }}
+        proxy_host: ${{ secrets.PROXY_HOST }}
+        proxy_username: ${{ secrets.USERNAME }}
+        proxy_key: ${{ secrets.SSH_PRIVATE_KEY }}
+        envs: DOCKER_CLIENT_TIMEOUT,COMPOSE_HTTP_TIMEOUT
+        script_stop: false
+        script: |
+          cd ${{ matrix.env}}
+          make integration
```
alexgarel, line 150: I personally don't think we should run integration test on the deployed server (for me integration tests are in the test pipeline…).
Most of the time integration tests are destructive, so you need to run them at least on a separate database for postgres / mongo etc.. and it's really risky to run them in a production environment.

If what you want to do is health check and or sanity checks we should use those names maybe.
At least we can have some test running in prod, but they should really be specific tests that are non destructive.

robotoff/settings.py
```diff
@@ -18,6 +18,10 @@ def _instance_tld() -> str:
         return _robotoff_instance
 
 
+_default_robotoff_doimain = f"openfoodfacts.{_instance_tld()}"
+_robotoff_domain = os.environ.get("ROBOTOFF_DOMAIN", _default_robotoff_doimain)
```
alexgarel, line 22:
```suggestion
_default_robotoff_domain = f"openfoodfacts.{_instance_tld()}"
_robotoff_domain = os.environ.get("ROBOTOFF_DOMAIN", _default_robotoff_domain)
```

tests/integration/test_connect.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.mark.integtest
+def test_connect_influxdb():
+    client = InfluxDBClient(
+        settings.INFLUXDB_HOST,
+        settings.INFLUXDB_PORT,
+        settings.INFLUXDB_USERNAME,
+        settings.INFLUXDB_PASSWORD,
+        settings.INFLUXDB_DB_NAME,
+    )
+    client.get_list_users()
```
alexgarel, line 36: Do we really need to use pytest to run this kind of things.

As already said (sorry for repeating) I would better have them as sanity checks, without any need of pytest to run this.

alexgarel (changes requested):
putting unit tests in their own folder is ok.

But the main problem that I see, is that those are sanity checks not integrations tests.

This is a really good thing to have, but we should distinguish from integration tests (destructives, to run in dev env) and we don't need pytest and so on to run those.

If we have those sanity checks we can maybe add the script running them to the "health_check" parameter of docker-compose

.env
alexgarel: seems like this one can be override on linux by adding:
```yaml
  extra_hosts:
    - host.docker.internal:host-gateway
```
on the services where you need it.

alexgarel: note: that it is a bit tricky though for the service you want to reach, needs to listen on the internal address corresponding to your docker sub network (this may be done by listening all interface, but even though it should listen live on newly created interfaces if your network is destroyed / recreated, like it happens on a down / up)

tests/integration/test_connect.py
alexgarel: Python is ok, I just mean we do not need pytest.

alexgarel: And also don't put them in tests/  which folder is not copied on prod image. Make a script directory or something like that.

ocervell: Not sure why pytest is not as appropriate as for unit tests for instance, it allows us to see individual results and know where it went wrong, compared to re-implementing those features in a non-pytest python-written test :/. 

For putting them in the `tests/` folder, that would allow to have `tests/unit`, `tests/integration`, `tests/health`, which seemed more consistent than having them in a separate folder

alexgarel: hum, ok this can be, but you will need to add pytest and all needed dependencies on prod image.

I'm also very reluctant to run integrations test (even unit tests) on production, as they can be destructive. So I would prefer the corresponding folders not to be copied on production image.

ocervell:
@alexgarel I've changed the way health checks are run:
* Health checks are now part of an API endpoint `/api/v1/health` that we can call after deployment (non-destructive).
* CI pipeline run `make up`, `make livecheck` and then `make health` to check connectivity with external services. 
* `make integration` is run only on .net environment.

alexgarel:
I worked on making running test locally possible (health check is not working but that would be difficult, as we yet don't have influx locally) and improve dev.

I did this not really on purpose (I though I was on master…) but it's useful work however.

I also have to continue work on this branch to avoid errors in preprod.

alexgarel (approved):
I think we are ready to merge.

PR #555, opened 2022-01-21
feat: make neural category apply automatically

alexgarel:
over a threshold

### DIFF
--- a/robotoff/prediction/types.py
+++ b/robotoff/prediction/types.py
@@ -67,7 +67,7 @@ class ProductPredictions:
 
     @classmethod
     def merge(cls, items: Iterable["ProductPredictions"]) -> "ProductPredictions":
-        """Merges predictions into one object, checking compatibility"""
+        """Merge predictions into one object, checking compatibility."""
         items = list(items)
         if len(items) == 0:
             raise ValueError("no items to merge")
--- a/robotoff/prediction/types.py
+++ b/robotoff/prediction/types.py
@@ -54,13 +54,20 @@ def dict_factory(*args, **kwargs):
 
 @dataclasses.dataclass
 class ProductPredictions:
+    """Represent a list of prediction on a single product for a single type"""
+
+    #: the prediction list
     predictions: List[Prediction]
+    #: product barcode
     barcode: str
+    #: type of prediction
     type: PredictionType
+    #: optional image
     source_image: Optional[str] = None
 
     @classmethod
     def merge(cls, items: Iterable["ProductPredictions"]) -> "ProductPredictions":
+        """Merges predictions into one object, checking compatibility"""
         items = list(items)
         if len(items) == 0:
             raise ValueError("no items to merge")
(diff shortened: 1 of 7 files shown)


alexgarel:
@mithridatea :

- I would like to wait to push new robotoff to prod before merging, and we will first test it in preprod
- I would like to add a feature to add insight in database if automatic_processing is true, even if insight already exists (but without automatic_processing)

robotoff/prediction/types.py
```diff
@@ -54,13 +54,20 @@ def dict_factory(*args, **kwargs):
...
+    #: product barcode
     barcode: str
+    #: type of prediction
     type: PredictionType
+    #: optional image
     source_image: Optional[str] = None
 
     @classmethod
     def merge(cls, items: Iterable["ProductPredictions"]) -> "ProductPredictions":
+        """Merges predictions into one object, checking compatibility"""
```
mithridatea, line 70: Merges -> Merge

robotoff/workers/tasks/product_updated.py
```diff
@@ -79,15 +79,24 @@ def update_insights(barcode: str, server_domain: str):
 
 
 def add_category_insight(barcode: str, product: JSONType, server_domain: str) -> bool:
+    """Tries to compute categories
```
mithridatea, line 82: `Tries to compute categories` -> `Tries to compute categories.`

mithridatea (changes requested):

tests/integration/insights/test_category_import.py
```diff
@@ -0,0 +1,148 @@
...
+from robotoff.insights.importer import InsightImporterFactory
+from robotoff.models import ProductInsight
+from robotoff.prediction.types import Prediction, PredictionType, ProductPredictions
+from robotoff.products import Product
+
+insight_id1 = "94371643-c2bc-4291-a585-af2cb1a5270a"
+barcode1 = "00001"
+
+
+@pytest.fixture(autouse=True)
```
mithridatea, line 14: I would prefer not using `autouse=True`, as currently this fixture is run for all unit tests. Better including it manually as it is only useful for this module.

tests/integration/insights/test_category_import.py
```diff
@@ -0,0 +1,146 @@
+import pytest
```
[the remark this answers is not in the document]
alexgarel: Those are more integration test than unit tests, this is why I did the distinction.

tests/integration/insights/test_category_import.py
alexgarel: I don't think so, it is only run for the test of this module.

See [the documentation](https://docs.pytest.org/en/latest/reference/fixtures.html#fixture-availability)
> Similarly, a test can also only be affected by an autouse fixture if that test is in the same scope that autouse fixture is defined in (see Autouse fixtures are executed first within their scope).

It would be true if fixture was in conftest.py

alexgarel: I tested it running with a pdb, to be sure.

mithridatea (approved):

PR #560, opened 2022-01-25
ops: avoid mangling org and net metrics

alexgarel:
by setting a different influxdb database (they use same instance)


docker-compose.yml
```diff
@@ -28,6 +28,7 @@ x-robotoff-base-env: &robotoff-base-env
   OFF_PASSWORD:
   INFLUXDB_HOST:
   INFLUXDB_PORT:
+  INFLUX_DB_NAME:
```
ocervell, line 31: This should be `INFLUXDB_NAME`

.github/workflows/container-deploy.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           echo "TF_SERVING_HOST=${{ secrets.TF_SERVING_HOST }}" >> .env
           echo "INFLUXDB_HOST=${{ secrets.INFLUXDB_HOST }}" >> .env
           echo "INFLUXDB_PORT=8086" >> .env
-          echo "INFLUXDB_NAME=db0" >> .env
+          echo "INFLUXDB_NAME=${{ env == 'robotoff-org' && 'off_metrics' || 'off_net_metrics' }}" >> .env
```
ocervell, line 104: I'm not sure `env == ` will work here, but `matrix.env ==` should ?

ocervell (changes requested):

.github/workflows/container-deploy.yml
alexgarel: thanks !

docker-compose.yml
alexgarel: seems I was tired while writing this :sweat_smile:

alexgarel:
@ocervell if you want to review again ! Thanks…

ocervell (approved):
LGTM

PR #563, opened 2022-01-26
fix: set num workers in gunicorn from env

alexgarel:
for we need to take into account memory consumption

**WARNING** This PR merges against v1.1.0

### DIFF
--- a/gunicorn.py
+++ b/gunicorn.py
@@ -1,9 +1,10 @@
-import multiprocessing
+import os
 
 from robotoff import models
 
 bind = ":5500"
-workers = multiprocessing.cpu_count() * 2 + 1
+# we have a trade-off with memory vs cpu numbers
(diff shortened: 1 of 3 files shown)


gunicorn.py
```diff
@@ -1,9 +1,11 @@
 import multiprocessing
+import os
 
 from robotoff import models
 
 bind = ":5500"
-workers = multiprocessing.cpu_count() * 2 + 1
+# we have a trade-off with memory vs cpu numbers
+workers = os.environ.get("GUNICORN_NUM_WORKERS", 4)
```
ocervell, line 8: casting to an int as env variables are loaded as strings.

```suggestion
workers = int(os.environ.get("GUNICORN_NUM_WORKERS", 4))
```

ocervell (approved):

gunicorn.py
alexgarel: oops ! right !

PR #570, opened 2022-01-28
fix: straight conditions for category insights

alexgarel:
This is a fix to avoid predicting categories when we should not.

@mithridatea you need to confirm my choice that we should not predict with neural if ingredients is empty (I may be wrong).

### DIFF
--- a/robotoff/insights/importer.py
+++ b/robotoff/insights/importer.py
@@ -341,6 +341,9 @@ class CategoryImporter(InsightImporter):
         category: str,
         seen_set: Set[str],
     ):
+        if product is None:
+            return False  # product not in store !
+
         product_categories_tags = getattr(product, "categories_tags", [])
 
         # first check whether this is new information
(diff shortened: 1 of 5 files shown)


mithridatea:
The model was trained on data with and without ingredients (see https://github.com/openfoodfacts/robotoff/blob/master/scripts/category_dataset.py#L109 for the filter used on the dataset that was used to trained the model), only the product name is a required field.

robotoff/insights/importer.py
```diff
@@ -341,6 +341,9 @@ def ignore_insight(
         category: str,
         seen_set: Set[str],
     ):
+        if product is None:
+            return False  # product not in store !
```
mithridatea, line 345: I forgot to inverse the bool, we should return `True` if we ignore the insight and `False` otherwise. Can you inverse all `return True/False` and remove the `not` on L332?

alexgarel: done

mithridatea (approved):

PR #592, opened 2022-02-02
Make `prediction.automatic_processing` field nullable

mithridatea:
As we further decouple prediction from insight import, we need to keep
track if the prediction automatic_processing is null or not.


alexgarel (approved):
:+1:

PR #616, opened 2022-02-09
Move OCR category predictor to legacy folder

mithridatea:
This predictor is currently not used, as it's not served by Tensorflow Serving. Move it to `legacy` folder until:
- we either convert it so it is served by Tensorflow Serving (or another ML serving service)
- we include OCR text as input in the category detector

Pros: faster unit tests, no need to download any model anymore


alexgarel (approved):
LGTM

PR #634, opened 2022-02-22
Semantic PR check

teolemon:
### What
- Semantic PR check

### DIFF
--- a/.github/workflows/semantic-pr.yml
+++ b/.github/workflows/semantic-pr.yml
@@ -1,0 +1,17 @@
+name: "Semantic PRs"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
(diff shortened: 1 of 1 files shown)


alexgarel (changes requested):
I think you have to configure types, for we don't want to be limited to feat/fix.

I suggest having at least:
- fix
- feat
- doc
- refactor
- ops

You will have to change crowdin commits message to be "feat: add translations" or something like it.

teolemon:
I believe this is the default: https://github.com/commitizen/conventional-commit-types/blob/master/index.json

teolemon:
ops is not included

alexgarel (approved):
ok, sorry, for the defaults, I misread the doc. We could also go for the default. (It's ok to change ops for build / ci).

As you prefer.

PR #658, opened 2022-03-20
feat(api): Question API, increase query randomness, and allow to use pagination

alexfauquette:
### What

This modification is proposed to provide more flexibility in question fetching especially for hunger-game

### Fixes bug(s)
We were relying on the randomness of question requests to populate the batch of questions. But now, you can run multiple time the random query and get the same list of questions
https://robotoff.openfoodfacts.org/api/v1/questions/random?count=10&lang=fr&insight_types=category&value_tag=en:teas

My assumption is that if no product has obtained a vote, `random() * nb_votes` is always 0 and so the order is the "natural" one in the database.

I also reused codes from insight API request to add `page` parameter to the questions route such that event with `popular` soring we are able to fetch all the data

### Part of 

https://github.com/openfoodfacts/openfoodfacts-hungergames/issues/394

### DIFF
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -166,6 +166,10 @@ class InsightCollection:
 
 class RandomInsightResource:
     def on_get(self, req: falcon.Request, resp: falcon.Response):
+        response: JSONType = {}
+        count: int = req.get_param_as_int("count", min_value=1, default=25)
+        page: int = req.get_param_as_int("page", min_value=1, default=1)
+        
         insight_type: Optional[str] = req.get_param("type")
         country: Optional[str] = req.get_param("country")
         value_tag: Optional[str] = req.get_param("value_tag")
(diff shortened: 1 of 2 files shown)


robotoff/app/core.py
```diff
@@ -111,7 +111,8 @@ def get_insights(
 
     if order_by is not None:
         if order_by == "random":
-            query = query.order_by((peewee.fn.Random() * ProductInsight.n_votes).desc())
+            # The +1 is here to avoid 0*rand() = 0
+            query = query.order_by((peewee.fn.Random() * (ProductInsight.n_votes + 1)).desc())
```
alexgarel, line 115: :clap: really good catch !

robotoff/app/api.py
```diff
@@ -166,26 +166,39 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
...
+            country=country,
+            server_domain=server_domain,
+            value_tag=value_tag,
+            order_by="random",
+            server_domain=server_domain,
         )
 
-        resp.media = {"insights": [insight.serialize() for insight in insights]}
+        offset: int = (page - 1) * count
+        insights = [i.serialize() for i in get_insights_(limit=count, offset=offset)]
```
alexgarel, line 191: Is there any value to add a page to a random query ? The random query will gives you different results every time, so pagination does not really have sense (but if we cached the results).

Maybe we can just ignore them (result is different every time), because taking them into account gonna make things less performant for no gain.

robotoff/app/core.py
alexgarel: BTW it made me think of adding #664 bug about performances problem of this piece of code.

alexgarel (changes requested):
Cool, but don't you agree we should ignore pagination for random questions ?

robotoff/app/api.py
alexfauquette: Yes, for random it does not make sens

robotoff/app/api.py
```diff
@@ -948,7 +962,8 @@ def get_questions_resource_on_get(
         avoid_voted_on=_get_skip_voted_on(auth, device_id),
     )
 
-    insights = list(get_insights_(limit=count))
+    offset: int = (page - 1) * count
```
alexfauquette, line 965: here we could add a condition, such that if `order_by=='random'` then `offset=0` don't know if it improves perf

alexfauquette (commented):
@alexgarel do you have enough bandwidth to clean this PR such that tests pass?

Not sure I will have time to install the project this week

robotoff/app/api.py
alexgarel: yes good idea

alexgarel:
I made a new PR #666, cherry picking your commits, as I imagine I can't push to your repo.

alexfauquette:
Thanks for this diabolical PR 😈 (PR number 666 😉)

PR #666, opened 2022-03-22
fix(api): Question API, fix query randomness, and allow to use pagination

alexgarel:
### What

This modification is proposed to provide more flexibility in question fetching especially for hunger-game

### Fixes bug(s)
We were relying on the randomness of question requests to populate the batch of questions. But now, you can run multiple time the random query and get the same list of questions
https://robotoff.openfoodfacts.org/api/v1/questions/random?count=10&lang=fr&insight_types=category&value_tag=en:teas

My assumption is that if no product has obtained a vote, `random() * nb_votes` is always 0 and so the order is the "natural" one in the database.

I also reused codes from insight API request to add `page` parameter to the questions route such that event with `popular` soring we are able to fetch all the data

### Part of 

https://github.com/openfoodfacts/openfoodfacts-hungergames/issues/394

### DIFF
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -166,6 +166,10 @@ class InsightCollection:
 
 class RandomInsightResource:
     def on_get(self, req: falcon.Request, resp: falcon.Response):
+        response: JSONType = {}
+        count: int = req.get_param_as_int("count", min_value=1, default=25)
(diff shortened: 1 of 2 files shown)


alexfauquette (approved):
Looks good to me, thanks again for the fix :)

Issue #623, opened 2022-02-14
Use OpenAPI to document Robotoff API


mithridatea:
Currently, the Robotoff API is partially documented in https://github.com/openfoodfacts/robotoff/blob/master/doc/references/api.md, in the Blueprint API format.

Switch API documentation to a more recent and widely adopted API format such as OpenAPI. Many renderer exists, such as redocly to render the API and play with it: https://redocly.github.io/redoc/

Example: https://redocly.github.io/redoc/?url=https://openfoodfacts.github.io/robotoff/api.yml

alexgarel:
It's about using a tool to generate .md files from the openapi description (which is in the api.yml if I'm correct). The tool have to be integrated to be run automatically on documentation deployment (this is a github action) to generate the file: in generate-docs.yml and in doc target in Makefile (though that would be cool to have generate-docs use make docs !)

alexgarel:
First step: find the right tools to generate open api (json or yaml) from the source code.

alexgarel:
see https://pypi.org/project/falcon-openapi/  maybe

alexgarel:
See https://github.com/falconry/falcon/wiki/Add-on-Catalog

> [falcon-apispec](https://github.com/alysivji/falcon-apispec) [APISpec](https://github.com/marshmallow-code/apispec) plugin that generates OpenAPI specification (aka Swagger) for Falcon web applications
> [py2swagger](https://github.com/Arello-Mobile/py2swagger) OpenAPI specification generator

raphael0202:
After discussion with @alexgarel, we're going to directly right the documentation in doc/references/api.yml, without using an automated method to generate the documentation from the source code.

raphael0202:
Done, OpenAPI API is now used: https://github.com/openfoodfacts/robotoff/blob/master/doc/references/api.yml

PR #673, opened 2022-03-25
feat : api.md updated to openAPI

ram690514:
### What
- description of the PR

### Screenshot
(Insert a screenshot to provide visual record of your changes, if visible)
### Fixes bug(s)
- #1, #2 and #3 (change by appropriate issues)
### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/1
(please be as granular as possible)

### DIFF
--- a/.github/workflows/semantic-pr.yml
+++ b/.github/workflows/semantic-pr.yml
@@ -18,13 +18,7 @@ jobs:
         with:
           # Configure which types are allowed.
           # Default: https://github.com/commitizen/conventional-commit-types
-          types: |
-            fix
-            feat
-            doc
-            refactor
-            ops
-            chore
+          types: feat
           # Configure which scopes are allowed.
           #scopes: |
            # core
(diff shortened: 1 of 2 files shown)


.github/workflows/semantic-pr.yml
```diff
@@ -18,13 +18,7 @@ jobs:
...
           # Configure which types are allowed.
           # Default: https://github.com/commitizen/conventional-commit-types
-          types: |
-            fix
-            feat
-            doc
-            refactor
-            ops
-            chore
+          types: feat
```
alexgarel, line 21: oops you should not change that ! Those are *all* the possible qualifiers to commit messages !

alexgarel:
@ram690514 
- please take care of your PR description, why are you referencing tickets that have nothing to do here ? You should link to https://github.com/openfoodfacts/robotoff/issues/623 instead
- you misunderstood what should be done: it's not about copying the data in `.md` file, but using a tool to genertae `.md` files from the openapi description (which is in the api.yml if I'm correct). The tool have to be integrated to be run automatically on documentation deployment (this is a github action) to generate the file (if there is a `make doc`, this is the right place).

PR #674, opened 2022-03-26
feat: add events requests

ocervell:

PR #677, opened 2022-03-30
feat: add events API requests

alexgarel:
see https://github.com/openfoodfacts/openfoodfacts-events

Replaces #674

### What

Call openfoodfacts-events each time a user annotate an item

### DIFF
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -20,7 +20,6 @@ from robotoff import settings
 from robotoff.app import schema
 from robotoff.app.auth import BasicAuthDecodeError, basic_decode
 from robotoff.app.core import SkipVotedOn, SkipVotedType, get_insights, save_annotation
-from robotoff.app.events import send_event_async
 from robotoff.app.middleware import DBConnectionMiddleware
 from robotoff.insights.extraction import (
     DEFAULT_OCR_PREDICTION_TYPES,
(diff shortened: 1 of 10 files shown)


alexgarel:
Hi @ocervell I built upon your commit, but I changed it a bit.
- added a test
- only launch one process to have async posting (I think it was a bit too much to have one process for each events, also because we have in-process cache, which can make the fork costy after a while) 
- in dev setup: no post to events (I think few dev will have an events project setup)
- changed location of event sending to be able to have the barcode

I would be glad if you want to review it.

robotoff/app/api.py
```diff
@@ -812,7 +812,8 @@ def on_post(self, req: falcon.Request, resp: falcon.Response):
         barcode = req.get_param("barcode", required=True)
         action = req.get_param("action", required=True)
         server_domain = req.get_param("server_domain", required=True)
-
+        print(server_domain)
+        print(settings.OFF_SERVER_DOMAIN)
```
ocervell, line 816: remove the print statements before merging

ocervell (approved):
LGTM, but still needs testing within Docker local env and in dev env after !

alexgarel:
Thanks for the review @ocervell 

I've done some test in local and it worked like a charm.

I will make a push to preprod to verify (through a specific PR on a deploy-* branch)

alexgarel:
Tested ok in preprod (#718)

Issue #692, opened 2022-04-09
Adding MongoDB docker to Robotoff


Jagrutiti:
### What
When trying to run Robotoff's `save_annotation` I encountered the error saying missing connection to MongoDB.
![Screenshot from 2022-04-08 19-55-47](https://user-images.githubusercontent.com/28809492/162572990-ad50fe02-08f1-46ee-a6b5-3aff5185a10f.png)

For now the only solutions we have is:

- install perl server and run it with docker, aka openfoodfacts-server (eats lot of memory)

- add a mongodb docker in the same network and configure robotoff to use it

Here is a reference to the [Slack discussion](https://openfoodfacts.slack.com/archives/CF0286BRQ/p1649429663590889?thread_ts=1649345138.680329&cid=CF0286BRQ) about the same.

### Steps to reproduce the behavior:
1. run `docker-compose up`  
2. Try to execute the function in [pdb](https://docs.python.org/3.8/library/pdb.html) with these parameters:
`core.save_annotation(insight_id='3e168efb-6c5d-4933-bb32-a928cd0dc0ae',annotation=1,device_id='57321fd2a838d4c11db6d7b0926c681d6ac5c543',update=True,data= None,auth= None,trusted_annotator = True)
`
4. See the execution breaking in the middle.

raphael0202:
As the PR is merged, I'm closing the issue.

Issue #606, opened 2022-02-04
AnnotationResult should be stored in database


alexgarel:
**Is your feature request related to a problem? Please describe.**

When robotoff processes an insight (process insight), an action may happen or not. Annotator.annotate return a useful AnnotationResult giving informations of what really happens : UPDATED_ANNOTATION_RESULT, MISSING_PRODUCT_RESULT, ALREADY_ANNOTATED_RESULT etc.

This is a very useful info when you want to know which insight did really have an impact. But currently this is not stored in the database.
This would be useful for stats and for debugging.

**Describe the solution you'd like**
Associate each AnnotationResult with a number and store it in `annotation` column (instead of a simple 1).

Alternatively, if this is not compatible with current code, We may store this info in another column.

alexgarel:
@teolemon annd @mithridatea you're opinion matters.

Jagrutiti:
Hi @alexgarel 

I would like to take this up. 

If I am not wrong this is the file in question
 https://github.com/openfoodfacts/robotoff/blob/master/robotoff/insights/annotate.py

Could you guide me at what point shall I save the data?

> Associate each AnnotationResult with a number and store it in `annotation` column (instead of a simple 1).

By this, you mean the PostgreSQL database? Would I need database dump or something? Because database setup instructions are not mentioned in the readme or contributing file.

alexgarel:
If you can run the tests, it means you have a database set up !

At first launch database is created, it is managed by peewee. See https://github.com/openfoodfacts/robotoff/blob/master/robotoff/models.py (you may also find database creation).

But at the moment we do not handle migrations (we have not many deployment, so we handle it manually…).

But you are right that we do not provide sample data and this is a problem, we should provide a sample dump of a few rows of each table. I will see if I can work on it, but it does not stops you.

alexgarel:
Btw a quick way to create some data is to use the factories in https://github.com/openfoodfacts/robotoff/blob/master/tests/integration/models_utils.py (in a python console that you can launch with poetry run python).

Jagrutiti:
Hey @alexgarel 

You were right about having a database setup while running test cases, also got the database structure. 

Got this URL: https://robotoff.openfoodfacts.org/api/v1/insights/ from Hunger games to fetch insight. 

From [api.md](https://github.com/openfoodfacts/robotoff/blob/master/doc/references/api.md#insight-annotations-insightsannotate) I understood to save we call `insights/annotate`

[...]

PR #693, opened 2022-04-09
feat: Adding MongoDB container to Robotoff in dev

Jagrutiti:
### What
I have tried to add the MongoDB container to docker.
When I run `docker-compose up` no error is thrown. So I think it works fine.
Screenshot from 2022-04-09 18-11-04

But when I call the `save_annotation` function through pdb it throws the following error. 
Screenshot from 2022-04-09 18-18-14

Before adding these new changes the error message was like so

Screenshot from 2022-04-08 19-55-47

### Closes
- #692
### Part of 
- I was actually trying to work on https://github.com/openfoodfacts/robotoff/issues/606 when I encountered this error.


alexgarel (changes requested):
You don't need the mongodb.yml file

Jagrutiti:
Hey @alexgarel 

I am just not sure how to provide a demo database to the user to download and extract?

docker/dev.yml
```diff
@@ -44,6 +44,12 @@ services:
   scheduler:
     <<: *networks-productopener-local
     <<: *robotoff-dev
+  mongodb:
+    # Note: keep it as close a possible to mongodb.yml
```
raphael0202, line 48: This line should be removed I think

doc/how-to-guides/deployment/dev-install.md
```diff
@@ -51,4 +51,37 @@ After cloning the repository:
...
+    ```
+    tar xzf off-dev-mongo-dump.tar.gz
+    docker cp -a dump robotoff_mongodb_1:/var/tmp/
+    docker exec mongodb mongorestore /var/tmp/dump
+
+    ```
+
+    First we have extracted the zip file. Next we copy the dump in our running MongoDB service in Robotoff `robotoff_mongodb_1`. Lastly, we restore the dump from Robotoff's service to `mongodb` container.
+
+    The MongoDB configuration is now compelete.
```
raphael0202, line 77: compelete -> complete

raphael0202 (approved):

PR #749, opened 2022-05-04
ops: don't run on dependabot made prs as it will fail

teolemon:
### What
- ci: don't run on dependabot made prs as it will fail


alexgarel (approved):
LGTM

Issue #115, opened 2020-02-25
Do not apply nutriscore prediction if it does not match the current nutriscore of the product.


teolemon:
Do not apply nutriscore prediction if it does not match the current nutriscore of the product. Send an alert to slack instead

alexgarel:
This one is a real bug.

Import happens through robotoff.insights.importer.LabelInsightImporter

I don't look closely but the best strategies would be to filter it out in LabelInsightImporter if we already have product data from openfoodfacts.
If not, we will have to wait for execution by the scheduler which will apply it to the database. See robotoff.insights.annotate.InsightAnnotator.

teolemon:
We need to put priorities, and triage Robotoff issues.

alexgarel:
@Jagrutiti : here is some data to craft your prediction to feed in as examples to import. It's for product [3302740030949](https://world.openfoodfacts.org/product/3302740030949/)

```
postgres=# select * from prediction where barcode = '3302740030949' and type='label' and value_tag like 'en:nutriscore-%';
   id    |    barcode    | type  |                                                                   data                                                                   |         timestamp          |    value_tag    | value |       source_image       | automatic_processing |     server_domain     | predictor 
---------+---------------+-------+------------------------------------------------------------------------------------------------------------------------------------------+----------------------------+-----------------+-------+--------------------------+----------------------+-----------------------+-----------
 1675804 | 3302740030949 | label | {"model": "nutriscore", "notify": true, "confidence": 0.871872, "bounding_box": [1441.3848, 1036.8024, 1685.6928, 1430.2488]}            | 2019-10-09 05:46:31.07158  | en:nutriscore-c |       | /330/274/003/0949/6.jpg  | f                    | api.openfoodfacts.org | 
 2029750 | 3302740030949 | label | {"model": "nutriscore", "notify": true, "confidence": 0.816114128, "bounding_box": [0.591395915, 0.477426708, 0.688566923, 0.624632716]} | 2020-04-08 05:44:08.851982 | en:nutriscore-a |       | /330/274/003/0949/22.jpg | f                    | api.openfoodfacts.org | 
[...]

PR #772, opened 2022-05-19
fix: Do not apply nutriscore prediction if it does not match the current nutriscore of the product

Jagrutiti:
### What
- This PR compares the predicted and current nutriscore.

### Why
 The Nutri-Score model is not very reliable

### Fixes bug(s)
- #115 
### Part of 
- https://github.com/openfoodfacts/openfoodfacts-server/issues/5551


tests/unit/workers/tasks/test_import_image.py
```diff
@@ -0,0 +1,22 @@
+from PIL import Image
+from robotoff.utils import get_image_from_url, get_logger, http_session
+
+image = get_image_from_url('https://world.openfoodfacts.org/images/products/802/509/314/0251/4.jpg', error_raise=False)
```
alexgarel, line 4: good idea, but you should put it in the repository instead (in a tests/data directory for example), we wont to avoid tests depending on an external service.

alexgarel: An image with a nutriscore would be more cool ;-) 
Like https://world.openfoodfacts.org/images/products/330/274/003/0949/front_fr.110.400.jpg

(There are plenty, see eg https://world.openfoodfacts.org/label/nutriscore-grade-c)

tests/unit/workers/tasks/test_import_image.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_import_insights_from_image(mocker):
+    barcode = "4640028621441"
+    source_image = 'https://world.openfoodfacts.org/images/products/802/509/314/0251/4.jpg'
+    ocr_url = 'https://world.openfoodfacts.org/images/products/802/509/314/0251/4.json'
+    sever_domain = 'http://openfoodfacts.org/'
+    
+    expected_predictions_all = get_predictions_from_image(barcode, image, source_image, ocr_url, sever_domain)
```
alexgarel, line 13: It's strange that you call it, to mock it ! 
In test you won't have a model.
Instead you should create a list of Prediction using PredictionFactory.

You are a bit lost maybe because you don't know what to put inside it, but what you are testing should tell you what to put in it. I can also provides you some data from production.

tests/unit/workers/tasks/test_import_image.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_import_insights_from_image(mock):
+    barcode = "3302740030949"
+    source_image = '/330/274/003/0949/6.jpg '
+    ocr_url = 'https://world.openfoodfacts.org/images/products/802/509/314/0251/4.json'
+    server_domain = 'api.openfoodfacts.org'
+
+    expected_predictions_all =  [
+        PredictionFactory(type='label',  barcode=3302740030949, data={"model": "nutriscore", "confidence": 0.871872, "bounding_box": [1441.3848, 1036.8024, 1685.6928, 1430.2488]}, value_tag="en:nutriscore-c", value="", automatic_processing="f"),
```
alexgarel, line 15: you should write: automatic_processing=False ("f" in postgres means False)

tests/unit/workers/tasks/test_import_image.py
```diff
@@ -1,27 +1,82 @@
...
 
-    # import_insights = mock.Mock(return_value=)
-    
+    mock_import_insights.assert_called_once_with(
+        predictions=predictions_list,
+        server_domain=server_domain,
+        automatic=True,
+    )
 
+    mock_get_predictions_from_image.assert_called_once_with(
```
Jagrutiti, line 80: Both `assert_called_once_with` throw the error saying `assert_called_once_with shou
d be called once but called 0 times`

I was not able to resolve it. I did see the existing examples in the code. I read [where to patch](https://docs.python.org/3/library/unittest.mock.html#where-to-patch) and tried to follow the steps in the document but no luck.

tests/unit/workers/tasks/test_import_image.py
```diff
@@ -1,27 +1,82 @@
...
 
-    get_predictions_from_image = mock.Mock(return_value=expected_predictions_all)
+    image = Image.open(requests.get(source_image, stream=True).raw)
 
-    get_product = {"code":"3302740030949","product":{"labels":"Sans colorants, Nutriscore, Nutriscore C, Fumé au bois de hêtre","product_name":"Pommes de terre gratinées aux lardons et champignons"},"status":1,"status_verbose":"product found"}
+    mock_import_insights = mocker.patch(
+        "robotoff.insights.importer.import_insights", return_value=4
+    )
+    mock_get_predictions_from_image = mocker.patch(
+        "robotoff.insights.extraction.get_predictions_from_image"
```
Jagrutiti, line 20: Looking at the code I was not able to figure out what will `get_predictions_from_image`  return

teolemon:
=========================== short test summary info ============================
FAILED tests/unit/workers/tasks/test_import_image.py::test_import_insights_from_image
================= 1 failed, 315 passed, 10 warnings in 27.20s ==================
1
make: *** [Makefile:143: unit-tests] Error 1
Error: Process completed with exit code 2.

alexgarel:
Bug was closed.

PR #796, opened 2022-06-08
feat: Robotoff quality monitoring: Saving AnnotationResult

Jagrutiti:
### What
Saving Annotation result returned by `Annotator.annotate`

### Closes
- #606


alexgarel:
The problem with this PR, is that I'm not sure if we can change annotation meaning. I fear not, because external tools might use it.
I think we should better go with a new field on the model. But note that this require modifying the database manually (we have no migration mechanism right now).

robotoff/insights/annotate.py
```diff
@@ -107,11 +108,12 @@ def _annotate(
...
 
-        insight.save()
-
         if annotation == 1 and update:
             return self.process_annotation(insight, data=data, auth=auth)
 
+        insight.annotated_result = 1
+        insight.save
+
         return SAVED_ANNOTATION_RESULT
```
alexgarel, line 117: several comments:

1. you have to call the method save ;-) --> `insight.save()`
2. by moving this line under the `if` where there is a return, you are changing code logic. So you have to move it back on top
3. why use "1", when what we want to save is the returned value ! here, SAVE_ANNOTATION_RESULT

I would write:

```suggestion
        insight.save()
        if annotation == 1 and update:
            annotation_result = self.process_annotation(insight, data=data, auth=auth)
        else:
            annotation_result = SAVED_ANNOTATION_RESULT
        insight.annotated_result = 1
        insight.save()  # second call to save annotated_result
        return annotation_result
```

alexgarel: Ok, reading what you wrote below I understand you changes. 

That said, it's really better to keep things centralized and avoid spread logic for a specific detail all around the code. So I think this proposed pattern is better (even if it requires two "save" call)

alexgarel (changes requested):
Can we also change some test to see if annotated_result is the one expected ?
In test_annotate_insight_authenticated at least ?

robotoff/insights/annotate.py
```diff
@@ -107,12 +107,15 @@ def _annotate(
...
-        insight.save()
-
         if annotation == 1 and update:
-            return self.process_annotation(insight, data=data, auth=auth)
+            annotation_result = self.process_annotation(insight, data=data, auth=auth)
+        else:
+            annotation_result = SAVED_ANNOTATION_RESULT
 
-        return SAVED_ANNOTATION_RESULT
+        insight.annotated_result = annotation_result
```
alexgarel, line 115: annotation_result is a AnnotationResult object, so it should be:

```suggestion
        insight.annotated_result = annotation_result.status
```
Although I'm not sure if using a string is best…

Ideally, we should modify AnnotationResult to a an "id" (int) field, which 
* would be set to the int representing the status (see AnnotationStatus enum). 
* For that we would need to change all AnnotationResult instanciation, to add this parameter.
* and of course we would need to use an int field

tests/integration/test_api.py
```diff
@@ -185,6 +185,16 @@ def test_annotate_insight_authenticated(client):
...
 
+    # to test if "annotated_result" is being saved in the database
+    insight = next(
+        ProductInsight.select()
+        .where(ProductInsight.annotated_result.is_null(False))
+        .dicts()
+        .iterator()
+    )
+    assert insight.items() > {"username": "a", "annotation": -1, "n_votes": 0}.items()
+    assert "completed_at" in insight
```
alexgarel, line 196: You already have the insght, the only thing you have to test is that annotated_result is the expected value !

```suggestion
    assert insight.completed_at == "updated"
```

raphael0202 (approved):

raphael0202:
I've added the column in DB in preprod and prod:
```
ALTER TABLE product_insight
ADD COLUMN annotated_result INTEGER NULL;
```

PR #805, opened 2022-06-14
feat: Test case to test if we need 3 votes (sent with user id) to get the question answered definitely 

### LINKED ISSUE #801: we still need 3 votes (sent with user id) to get the question answered definitely
> mmmh… It was one hour ago, and didn't give you credit: https://world.openfoodfacts.org/product/2817896019485/st-nectaire-aop-laitier

The answers are resent in the end of questionnaire (after login) with user id.
I don't know how it works on the server side,  you mean if we send an answer with user id, we do not need to have the other votes? I am not sure about it ... since I tried with my account, we still need 3 votes (sent with user id) to get the question answered definitely (to have those scores )...

_Originally posted by @cli1005 in https://github.com/openfoodfacts/smooth-app/issues/2117#issuecomment-1143568155_

Jagrutiti:
### What
- Writing a test case to check if the authenticated user still needs votes to get the question answered.

tests
- #801

### DIFF
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -387,8 +387,9 @@ def test_annotation_event(client, monkeypatch, httpserver):
         )
     assert result.status_code == 200
 
+
 def test_annotate_insight_with_user_id(client):
-    # to test annotation insight when the user is authenticated and less votes are provided
+    # to test annotation insight when the user is authenticated with less votes are provided
     result = client.simulate_post(
         "/api/v1/insights/annotate",
         params={
@@ -401,7 +402,6 @@ def test_annotate_insight_with_user_id(client):
     assert result.status_code == 200
     assert result.json == {"description": "the annotation was saved", "status": "saved"}
 
-    
     authenticated_result = client.simulate_post(
         "/api/v1/insights/annotate",
         params={
@@ -413,7 +413,10 @@ def test_annotate_insight_with_user_id(client):
     )
 
     assert authenticated_result.status_code == 200
-    assert authenticated_result.json == {"description": "the annotation was saved", "status": "saved"}
+    assert authenticated_result.json == {
+        "description": "the annotation was saved",
+        "status": "saved",
+    }
     # For authenticated users we expect the insight to be validated directly, tracking the username of the annotator.
     votes = list(AnnotationVote.select())
     assert len(votes) == 0
@@ -426,7 +429,3 @@ def test_annotate_insight_with_user_id(client):
     )
     assert insight.items() > {"username": "a", "annotation": -1, "n_votes": 0}.items()
     assert "completed_at" in insight
-
-
-
-
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -433,9 +433,7 @@ def test_annotate_insight_anonymous_then_authenticated(client):
             "annotation": 1,
             "device_id": "voter1",
         },
-        headers={
-            "Authorization": "Basic " + base64.b64encode(b"a:b").decode("ascii")
-        }
+        headers={"Authorization": "Basic " + base64.b64encode(b"a:b").decode("ascii")},
     )
 
     assert authenticated_result.status_code == 200
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -386,3 +386,47 @@ def test_annotation_event(client, monkeypatch, httpserver):
             },
         )
     assert result.status_code == 200
+
+def test_annotate_insight_with_user_id(client):
+    # to test annotation insight when the user is authenticated and less votes are provided
+    result = client.simulate_post(
+        "/api/v1/insights/annotate",
+        params={
+            "insight_id": insight_id,
+            "annotation": -1,
+            "device_id": "voter1",
+        },
+    )
+
+    assert result.status_code == 200
+    assert result.json == {"description": "the annotation was saved", "status": "saved"}
+
+    
+    authenticated_result = client.simulate_post(
+        "/api/v1/insights/annotate",
+        params={
+            "insight_id": insight_id,
+            "annotation": -1,
+            "device_id": "voter1",
+        },
+        headers={"Authorization": "Basic " + base64.b64encode(b"a:b").decode("ascii")},
+    )
+
+    assert authenticated_result.status_code == 200
+    assert authenticated_result.json == {"description": "the annotation was saved", "status": "saved"}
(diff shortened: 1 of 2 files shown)


tests/integration/test_api.py
```diff
@@ -386,3 +386,47 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    votes = list(AnnotationVote.select())
+    assert len(votes) == 0
+
+    insight = next(
+        ProductInsight.select()
+        .where(ProductInsight.id == insight_id)
+        .dicts()
+        .iterator()
+    )
+    assert insight.items() > {"username": "a", "annotation": -1, "n_votes": 0}.items()
```
Jagrutiti, line 430: I believe I should increase the count of `"n_votes"` to 1 because the user has already voted once

tests/integration/test_api.py
```diff
@@ -386,3 +386,46 @@ def test_annotation_event(client, monkeypatch, httpserver):
             },
         )
     assert result.status_code == 200
+
+
+def test_annotate_insight_with_user_id(client):
```
alexgarel, line 391:
```suggestion
def test_annotate_insight_anonymous_then_authenticated(client):
```

alexgarel (changes requested):
You're in the right direction I added comments.

tests/integration/test_api.py
```diff
@@ -422,30 +422,36 @@ def test_annotate_insight_anonymous_then_authenticated(client):
...
 
-    assert authenticated_result.status_code == 200
-    assert authenticated_result.json == {
-        "description": "the annotation was saved",
-        "status": "saved",
-    }
-    # We have the previous vote, but the last request should validate the insight directly
-    votes = list(AnnotationVote.select())
-    assert len(votes) == 1  # this is the previous vote
+    if insight["process_after"] != None:
```
Jagrutiti, line 426: `make checks` fails here with the message: 

`./tests/integration/test_api.py:426:33: E711 comparison to None should be 'if cond is not None:'`

alexgarel: But with this, test passes just because insight["process_after"] is None ! So the rest of the code is skipped !
Why did you put that. 
It should instead check that process_after is empty.
You can do this by adding it to assert at line 421:

```python
assert not any(insight[key] for key in ("username", "completed_at", "annotation", "process_after"))
```

tests/integration/test_api.py
```diff
@@ -422,30 +422,36 @@ def test_annotate_insight_anonymous_then_authenticated(client):
...
+        insight = next(
+            ProductInsight.select()
+            .where(ProductInsight.id == insight_id)
+            .dicts()
+            .iterator()
+        )
+        assert (
+            insight.items() > {"username": "a", "annotation": 1, "n_votes": 1}.items()
+        )
+        assert "completed_at" in insight
```
alexgarel, line 457:
```suggestion
# we still have the vote, but we also have an authenticated validation
assert insight.items() > {"username": "a", "n_votes": 1}.items()
# process after is not None so that it would be picked by scheduler process_insight
assert insight["process_after"] is not None
```

alexgarel (changes requested):
I commented, you have to remove your `if`

note: if your test is failing it also might be due to a bug in robotoff. Remember we are writing this test to verify if the bug is there.

tests/integration/test_api.py
```diff
@@ -423,35 +423,31 @@ def test_annotate_insight_anonymous_then_authenticated(client):
...
 
-    if insight["process_after"] != None:
-        authenticated_result = client.simulate_post(
-            "/api/v1/insights/annotate",
-            params={
-                "insight_id": insight_id,
-                "annotation": 1,
-                "device_id": "voter1",
-            },
-            headers={
```
Jagrutiti: I have removed `headers` here because it was trying to connect to MongoDB and I was unable to progress with testing at all. I tried to [get help from the internet](https://github.com/falconry/falcon/discussions/1969#discussioncomment-1465909) but their syntax was the same as ours. I wasn't able to figure out what to do next.

tests/integration/test_api.py
```diff
@@ -423,35 +423,31 @@ def test_annotate_insight_anonymous_then_authenticated(client):
...
+    assert len(votes) == 2  # this is the previous vote
+
+    insight = next(
+        ProductInsight.select()
+        .where(ProductInsight.id == insight_id)
+        .dicts()
+        .iterator()
+    )
+    # we still have the vote, but we also have an authenticated validation
+    assert insight.items() > {"username": "a", "n_votes": 2}.items()
```
Jagrutiti, line 451: I changed `n_votes` to 2 because we are voting twice it is only logical that the number of votes is 2.

Also, the test fails in the end. I am unsure if it meets the expected scenario (we wanted it to fail anyway) and if the test case is correct.  

I can really make use of your review here @alexgarel

tests/integration/test_api.py
```diff
@@ -386,3 +386,68 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    assert len(votes) == 1
+
+    insight = next(
+        ProductInsight.select()
+        .where(ProductInsight.id == insight_id)
+        .dicts()
+        .iterator()
+    )
+
+    assert not any(insight[key] for key in ("username", "completed_at", "annotation"))
```
alexgarel, line 421: Maybe I missed it but you should have added "process_after" here to the tuple

```suggestion
    assert not any(insight[key] for key in ("username", "completed_at", "annotation", "process_after"))
```

tests/integration/test_api.py
```diff
@@ -386,3 +386,68 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+
+    # then the user connects and vote for same insights
+
+    authenticated_result = client.simulate_post(
+        "/api/v1/insights/annotate",
+        params={
+            "insight_id": insight_id,
+            "annotation": 1,
+            "device_id": "voter1",
+        },
```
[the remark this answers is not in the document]
Jagrutiti: Hey @alexgarel 

I have made all the changes you reviewed. 

I could not run the test case after adding headers for authorization because it tried to connect to MongoDB :(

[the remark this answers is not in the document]
alexgarel: I also add to mock another function, but now it's passing.

alexgarel:
@Jagrutiti I will let you look at my modifications and approve them.

Note that this means that we didn't reproduce the bug. So I will comment on the bug.

Issue #618, opened 2022-02-11
Display insight / predictions etc for a product or a particular filter


alexgarel:
# What
- It's quite difficult to know every predictions / insight that robotoff made for a product, its images included

# Proposed solution

- I would like to have a simple url  giving all the predictions / insights / annotations that robotoff has about a product. 
Also which were applied, when etc.
- This should be geared towards analyze by a human to understand what robotoff did, for debugging / auditing etc.
- Currently in hungergames we display the insights : (at https://hunger.openfoodfacts.org/insights)

<img width="1437" alt="image" src="https://user-images.githubusercontent.com/1689815/172192784-ff6d731f-9fd8-47ec-bf63-f8d9be107d1f.png">

But we would like to add predictions and all things related to images.

Add apis akin to `/api/v1/insights` (ProductInsightResource), with compatible filters (on product and type) for:
1. Predictions
2. Images
3. ImagePrediction - with info about image
4. LogoAnnotation - with info about image and annotation

The idea would be to use those APIs to append more tables in hunger game to have a complete view of why robotoff is predicting something.

Also add those filters to specific views:
* for Images filter out images that have at least one ImagePrediction
* for ImagePrediction filter out the one that have a corresponding logo annotation

Those filters will avoid repeating informations that are already in ImagePrediction or LogoAnnontation in hunger game.

- Related ticket for the Hunger Games part: https://github.com/openfoodfacts/openfoodfacts-hungergames/issues/415

ram690514:
yeah, its good to have an html page that shows all predictions/insights . It could be also used to show the recent updates and changes in the system

ram690514:
I am interested to work on this issue, if no one is currently working on it

alexgarel:
No let's go @ram690514 I think it's a good issue to start with !

ashish-reddy-20-08:
Hey @alexgarel , is ram working on this, if not I can take it up, I will be able to manage all the insights and prediction stuff, just would need help at the last to post it in HTML.

alexgarel:
I don't think @ram690514 is currently working on this.

I think the best is to have the data as json + as html on another endpoint (the html view can just exploit the same method to get data and build the html)

Here are the point from most difficult to list difficult (you can start doing them incrementally and submit the PR as draft).

For the json:
1. list all ProductInsight for this barcode - sort them in a predictable way : type - timestamp - id
2. list all Prediction for this barcode (same sorting order)
[...]

PR #815, opened 2022-06-26
feat: display insight/prediction for a product

Jagrutiti:
Working on displaying insight/prediction etc to a product. 

Following the steps listed in the description.
Currently: list all ProductInsight for this barcode - sort them in a predictable way : type - timestamp - id

### Relates to
- #618


teolemon:
def test_popular_question_pagination(client, mocker):
        mocker.patch("robotoff.insights.question.get_product", return_value={})
        ProductInsight.delete().execute()  # remove default sample
        for i in range(0, 12):
            ProductInsightFactory(barcode=i, unique_scans_n=100 - i)
        result = client.simulate_get("/api/v1/questions/popular?count=5&page=1")
        data = result.json
        assert data["count"] == 12
        assert data["status"] == "found"
>       assert [q["barcode"] for q in data["questions"]] == ["0", "1", "2", "3", "4"]
E       AssertionError: assert ['11', '10', '9', '8', '7'] == ['0', '1', '2', '3', '4']
E         At index 0 diff: '11' != '0'
E         Full diff:
E         - ['0', '1', '2', '3', '4']
E         + ['11', '10', '9', '8', '7']
tests/integration/test_api.py:102: AssertionError
---------------------------- Captured stdout setup ---------

robotoff/app/core.py
```diff
@@ -126,6 +126,8 @@ def get_insights(
         elif order_by == "n_votes":
             query = query.order_by(ProductInsight.n_votes.desc())
 
+    query = query.order_by(ProductInsight.timestamp.desc())
+
```
alexgarel, line 130: you should add this only if we have a order_by == "most_recent" parameter, not always, because you are going to change existing code in an unpredictable manner.

This is what makes the test failed by the way (because your line removes previous ordering).

robotoff/app/api.py
```diff
@@ -1055,6 +1055,26 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
...
+            )
+        ]
+
+        if not insights:
+            response["status"] = "no_insights"
+        else:
+            response["insights"] = insights
+            response["status"] = "found"
+
+        resp.media = response
```
alexgarel, line 1075: Sorry for the ticket title was misleading.

Finally we decided to add api points akin to the one already existing for insights which is in InsightCollection

So what you have to do, is indeed make classes similar to InsightCollection (taking exactly same parameters, so you should try to refactor the parameter parsing code in a method or function), but returning predictions, images, etc.

This also probably mean you will have to do the same as get_insights function but for predictions, images etc.

Note: first do it for predictions, we shall see then after.

alexgarel (changes requested):
Here is my first review.

robotoff/app/api.py
Jagrutiti: Hi @alexgarel 
Made a new commit to get predictions.

robotoff/app/api.py
```diff
@@ -1055,6 +1055,51 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
...
+        brands = req.get_param_as_list("brands") or None
+        server_domain: Optional[str] = req.get_param("server_domain")
+
+        if keep_types:
+            # Limit the number of types to prevent slow SQL queries
+            keep_types = keep_types[:10]
+
+        if brands is not None:
+            # Limit the number of brands to prevent slow SQL queries
+            brands = brands[:10]
```
alexgarel, line 1076: I think we can factor this code between InsightCollection and DisplayInsightPredictionForProducts.

The idea, is that the function can return a dict (say `query_params`), and then simply use:
`get_predictions = functools.partial(get_predictions, **query_params)`

robotoff/app/api.py
```diff
@@ -1055,6 +1055,51 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
         resp.media = {"count": {"annotations": annotation_count}}
 
 
+class DisplayInsightPredictionForProducts:
```
alexgarel, line 1058: you have to add a url (a route) for this new class at the end of the file)

robotoff/app/api.py
```diff
@@ -1055,6 +1061,45 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
...
```
alexgarel, line 1064: Oh by the way, you should not call it `DisplayInsightPredictionForProducts` but just `PredictionCollection`, to be same as `InsightCollection`

robotoff/app/api.py
```diff
@@ -1106,3 +1151,6 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
 api.add_route("/api/v1/health", HealthResource())
 api.add_route("/api/v1/dump", DumpResource())
 api.add_route("/api/v1/users/statistics/{username}", UserStatisticsResource())
+api.add_route(
+    "/api/v1/insights/predictions/display", DisplayInsightPredictionForProducts()
+)
```
alexgarel, line 1156: Why not just `"/api/v1/insights/predictions"` to be coherent with the url for insights ?

robotoff/app/core.py
```diff
@@ -125,13 +125,39 @@ def get_insights(
...
+    if value_tag:
+        where_clauses.append(Prediction.value_tag == value_tag)
+
+    query = Prediction.select()
+
+    if where_clauses:
+        query = query.where(*where_clauses)
+
+    return query.iterator()
```
alexgarel, line 158: depending on the "count" parameter, you should not return this, look at get_insights to see how they do it.
```suggestion
    if count:
         return query.count()
    else:
        return query.iterator()
```

alexgarel (changes requested):
Here is to unblock you.

alexgarel (changes requested):
We have to add more tests:

- Right now your test is just testing status_code, it should check response content
- also you should then add a test where you create some predictions (PredictionFactory will help you do that) and check that it returns predictions to you
- Also you should check some filters to see they are working
- You should also add some tests on get_predictions (there again using PredictionFactory)

robotoff/app/api.py
```diff
@@ -1055,6 +1061,45 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
...
+
+        if brands is not None:
+            # Limit the number of brands to prevent slow SQL queries
+            brands = brands[:10]
+
+        query_parameters = {
+            "server_domain": server_domain,
+            "value_tag": value_tag,
+            "barcode": barcode,
+        }
```
alexgarel, line 1082: Good, but what about the other parameters : value_tag, brands

robotoff/app/api.py
```diff
@@ -1055,6 +1061,45 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
         resp.media = {"count": {"annotations": annotation_count}}
 
 
+class PredictionCollection:
+    def on_get(self, req: falcon.Request, resp: falcon.Response):
+        response: JSONType = {}
+        page: int = req.get_param_as_int("page", min_value=1, default=1)
+        count: int = req.get_param_as_int("count", min_value=1, default=25)
+        barcode: Optional[str] = req.get_param("barcode")
+        value_tag: str = req.get_param("value_tag")
```
alexgarel, line 1070: I think keep_types should also be pertinent here (Prediction type)

tests/integration/test_api.py
```diff
@@ -383,6 +389,39 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    result = client.simulate_get("/api/v1/insights/predictions")
+    assert result.status_code == 200
+    data = result.json
+    assert data["count"] == 2
+    assert data["status"] == "found"
+    predictions_data = data["predictions"]
+    assert predictions_data[0]["type"] == "category"
+    assert predictions_data[1]["type"] == "category"
+    assert predictions_data[0]["value_tag"] == "en:seeds"
+    assert predictions_data[1]["value_tag"] == "en:seeds"
```
Jagrutiti, line 418: There were other parameters like `timestamps`, `id`, `data` etc. But I did not use them because they were changing every time the factory was called. And a few other parameters were either `{}` or `None`

tests/integration/test_api.py
```diff
@@ -383,6 +389,39 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+
+
+def test_get_predictions():
```
Jagrutiti, line 421: I have added a test case for `get_predictions` as well. But I am unsure if I am covering all the filters you had in mind. 

Let me know if my approach is correct or you expected something else.

Also, 4 other test cases are failing which is weird I think because I did not touch the existing function (wrote new functions and test cases). I did find their reason for failing using PDB. But I think it is best you have a look first lest my fixing makes more mess.

robotoff/app/core.py
```diff
@@ -150,6 +152,12 @@ def get_predictions(
     if value_tag:
         where_clauses.append(Prediction.value_tag == value_tag)
 
+    if brands:
+        where_clauses.append(Prediction.brands == brands)
```
alexgarel, line 156: There is no brands on Prediction model, so I think you drop this brands parameter (also in the view).

tests/integration/test_api.py
alexgarel: yes, it's ok not to test every fields in a test. You just have to test, what proves you your test is right.

The id parameter might be important to test. To get it, your must do on line 397 and 408:
```
prediction1 = PredictionFactory(value_tag="en:seeds")
....
prediction2 = PredictionFactory(value_tag="en:seeds")
```
this gives you the created object, so that later on you can do:

```
assert prediction_data[0]["prediction.id"] = prediction1.id
...
assert prediction_data[1]["prediction.id"] = prediction2.id
```

tests/integration/test_api.py
```diff
@@ -383,6 +389,39 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    assert result.status_code == 200
+    data = result.json
+    assert data["count"] == 1
+    assert data["status"] == "found"
+    predictions_data = data["predictions"]
+    assert predictions_data[0]["barcode"] == "123"
+    assert predictions_data[0]["type"] == "category"
+    assert predictions_data[0]["value_tag"] == "en:seeds"
+
+    PredictionFactory(value_tag="en:seeds")
```
alexgarel, line 408: you should put different values for different attributes here to differentiate and see it works differently.
For example 
```
PredictionFactory(value_tag="en:beers", data={"sample": 1}, type="brand")
```

tests/integration/test_api.py
```diff
@@ -383,6 +389,39 @@ def test_annotation_event(client, monkeypatch, httpserver):
     assert result.status_code == 200
 
 
-def test_display_insight_prediction_for_products(client, mocker):
+def test_insight_prediction(client, mocker):
```
alexgarel, line 392: the name is not really descriptive.
More over you don't need the mocker parameter if you do not use it in your test.

```suggestion
def test_prediction_collection_no_filter(client):
```

robotoff/app/api.py
```diff
@@ -1106,3 +1160,4 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
 api.add_route("/api/v1/health", HealthResource())
 api.add_route("/api/v1/dump", DumpResource())
 api.add_route("/api/v1/users/statistics/{username}", UserStatisticsResource())
+api.add_route("/api/v1/insights/predictions/", PredictionCollection())
```
alexgarel, line 1163: Oh I didn't notice the url, early on. I think `/api/v1/predictions`  would be better
```suggestion
api.add_route("/api/v1/predictions/", PredictionCollection())
```

tests/integration/test_api.py
```diff
@@ -383,6 +389,39 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    assert predictions_data[1]["value_tag"] == "en:seeds"
+
+
+def test_get_predictions():
+    PredictionFactory(barcode="123")
+    actual = get_predictions(barcode="123")
+    actual_items = [item.to_dict() for item in actual]
+    assert actual_items[0]["barcode"] == "123"
+    assert actual_items[0]["type"] == "category"
+    assert actual_items[0]["value_tag"] == "en:seeds"
```
alexgarel, line 427: you could create at list 4 Prediction before starting with values varying for type, value_tag, barcode. Then test that using filters you get the one you want.

You can verify filter works by only verifying the ids you get.

Eg.: 
prediction1 --> barcode=123, type="category", value_tag="en:seeds"
prediciton2 --> barcode=123, type="category", value_tag="en:beers"
prediction3 --> barcode=123, type="label", value_tag="en:eu-organic"
prediction4 --> barcode=456, type="label", value_tag="en:eu-organic"

If you filter on barcode 123, you get the first 3,
if you filter on type "category" you get the first 2,
if you filter on value_tag "en:seeds", you only get the first,
if you filter on type="brand", you get none

Then you can try to combine filters like barcode=123 and type="category"

alexgarel (commented):
some instructions to go on !

tests/integration/test_api.py
```diff
@@ -420,29 +420,31 @@ def test_prediction_collection_no_filter(client):
...
 
     actual_prediction1 = get_predictions(barcode="123")
     actual_items1 = [item.to_dict() for item in actual_prediction1]
     assert actual_items1[0]["barcode"] == "123"
     assert actual_items1[0]["type"] == "category"
     assert actual_items1[0]["value_tag"] == "en:seeds"
+    assert len(actual_items1) == 1
 
-    actual_prediction2 = get_predictions(keep_types="category")
+    actual_prediction2 = get_predictions(keep_types="brand")
```
Jagrutiti, line 435: I cannot pass the `keep_types` parameter to `get_predictions`. It throws an error in the query.

alexgarel:
```suggestion
    actual_prediction2 = get_predictions(keep_types=["brand"])
```

tests/integration/test_api.py
```diff
@@ -381,3 +387,64 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    assert prediction_data[1]["id"] == prediction2.id
+    assert prediction_data[1]["type"] == "brand"
+    assert prediction_data[1]["value_tag"] == "en:beers"
+
+
+def test_get_predictions():
+    PredictionFactory(barcode="123", keep_types="category", value_tag="en:seeds")
+    PredictionFactory(type="category")
+    PredictionFactory(value_tag="en:seeds")
+    PredictionFactory(type="brand")
```
alexgarel, line 426: I would like those creations to reflect the data proposal here : https://github.com/openfoodfacts/robotoff/pull/815#discussion_r923060687

More over it's better to keep created predictions in variables (prediction1, prediction2…), so that you can check for ids in you test (first request you assert the fields, but then you just assert ids).

tests/integration/test_api.py
```diff
@@ -381,3 +387,82 @@ def test_annotation_event(client, monkeypatch, httpserver):
             },
         )
     assert result.status_code == 200
+
+
+def test_prediction_collection_no_filter(client):
+    result = client.simulate_get("/api/v1/predictions/")
+    assert result.status_code == 200
+    assert result.json == {"count": 0, "predictions": [], "status": "no_predictions"}
```
alexgarel, line 395: As this part of the test is independant from the rest, it's better to separate it.

```suggestion
def test_prediction_collection_no_result(client):
    result = client.simulate_get("/api/v1/predictions/")
    assert result.status_code == 200
    assert result.json == {"count": 0, "predictions": [], "status": "no_predictions"}

def test_prediction_collection_no_filter(client):
```

tests/integration/test_api.py
```diff
@@ -420,31 +420,49 @@ def test_prediction_collection_no_filter(client):
...
+    assert actual_items1[1]["value_tag"] == "en:beers"
+    assert actual_items1[2]["value_tag"] == "en:eu-organic"
+    assert len(actual_items1) == 3
 
     actual_prediction2 = get_predictions(keep_types=["brand"])
     actual_items2 = [item.to_dict() for item in actual_prediction2]
-    assert actual_items2[0]["barcode"] == "123"
-    assert actual_items2[0]["type"] == "brand"
-    assert actual_items2[0]["value_tag"] == "en:seeds"
+    assert prediction2.id == prediction1.id + 1
```
Jagrutiti, line 451: @alexgarel 

I am not sure about this line though

alexgarel: you should not have to test that. First it's not related to your test (you are not supposed to be testing PredictionFactory here), and second it's not garanteed to be true.

tests/integration/test_api.py
```diff
@@ -381,3 +387,85 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    assert actual_items1[0]["type"] == "category"
+    assert actual_items1[0]["value_tag"] == "en:seeds"
+    assert actual_items1[1]["value_tag"] == "en:beers"
+    assert actual_items1[2]["value_tag"] == "en:eu-organic"
+    assert len(actual_items1) == 3
+
+    actual_prediction2 = get_predictions(keep_types=["brand"])
+    actual_items2 = [item.to_dict() for item in actual_prediction2]
+    assert prediction2.id == prediction1.id + 1
+    assert len(actual_items2) == 0
```
alexgarel, line 452: No need to transform if it's an empty list !

```suggestion
    assert actual_items2 == []
```

tests/integration/test_api.py
```diff
@@ -381,3 +387,85 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    actual_items1 = [item.to_dict() for item in actual_prediction1]
+    assert actual_items1[0]["id"] == prediction1.id
+    assert actual_items1[0]["barcode"] == "123"
+    assert actual_items1[0]["type"] == "category"
+    assert actual_items1[0]["value_tag"] == "en:seeds"
+    assert actual_items1[1]["value_tag"] == "en:beers"
+    assert actual_items1[2]["value_tag"] == "en:eu-organic"
+    assert len(actual_items1) == 3
+
+    actual_prediction2 = get_predictions(keep_types=["brand"])
```
alexgarel, line 449: It's cool to add some descriptive comment, so that readers don't have to process all lines to understand (I let you do it on other parts)

```suggestion
    # test that as we have no "brand" prediction, returned list is empty
    actual_prediction2 = get_predictions(keep_types=["brand"])
```

tests/integration/test_api.py
```diff
@@ -381,3 +387,85 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    )
+
+    actual_prediction1 = get_predictions(barcode="123")
+    actual_items1 = [item.to_dict() for item in actual_prediction1]
+    assert actual_items1[0]["id"] == prediction1.id
+    assert actual_items1[0]["barcode"] == "123"
+    assert actual_items1[0]["type"] == "category"
+    assert actual_items1[0]["value_tag"] == "en:seeds"
+    assert actual_items1[1]["value_tag"] == "en:beers"
+    assert actual_items1[2]["value_tag"] == "en:eu-organic"
```
alexgarel, line 446: Here, to be more precise, you should also assert "id".

Do it also in tests below.

```suggestion
    assert actual_items1[1]["value_tag"] == "en:beers"
    assert actual_items1[1].id == prediction2.id
    assert actual_items1[2]["value_tag"] == "en:eu-organic"
    assert actual_items1[2].id == prediction3.id
```

tests/integration/test_api.py
```diff
@@ -381,3 +387,85 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    assert prediction2.id == prediction1.id + 1
+    assert len(actual_items2) == 0
+
+    actual_prediction3 = get_predictions(value_tag="en:eu-organic")
+    actual_items3 = [item.to_dict() for item in actual_prediction3]
+    assert actual_items3[0]["id"] == prediction3.id
+    assert actual_items3[0]["barcode"] == "123"
+    assert actual_items3[0]["type"] == "category"
+    assert actual_items3[0]["value_tag"] == "en:eu-organic"
+    assert len(actual_items3) == 2
```
alexgarel, line 460: you assert you have 2 items, but you do not test the second item value… (checking the id is enough)

```suggestion
assert len(actual_items3) == 2
assert actual_items3[1].id == prediction4.id
```

tests/integration/test_api.py
```diff
@@ -381,3 +387,85 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+
+    actual_prediction1 = get_predictions(barcode="123")
+    actual_items1 = [item.to_dict() for item in actual_prediction1]
+    assert actual_items1[0]["id"] == prediction1.id
+    assert actual_items1[0]["barcode"] == "123"
+    assert actual_items1[0]["type"] == "category"
+    assert actual_items1[0]["value_tag"] == "en:seeds"
+    assert actual_items1[1]["value_tag"] == "en:beers"
+    assert actual_items1[2]["value_tag"] == "en:eu-organic"
+    assert len(actual_items1) == 3
```
alexgarel, line 447: That's important to check that, but you should also put it as you first assert (before checking individual elements).

This is because otherwise, if you have 2 elements, your test will fail at line `assert actual_items1[2]["value_tag"] == "en:eu-organic"` with an `IndexOutOfRange` error instead of providing a nice assertion error.
If you first test the number of elements, you know you can check the third element thereafter.

alexgarel (changes requested):
Great @Jagrutiti, we are almost there, some small changes and it will be ok.

tests/integration/test_api.py
```diff
@@ -381,3 +387,91 @@ def test_annotation_event(client, monkeypatch, httpserver):
...
+    prediction2 = PredictionFactory(
+        value_tag="en:beers", data={"sample": 1}, type="brand"
+    )
+    result = client.simulate_get("/api/v1/predictions/")
+    assert result.status_code == 200
+    data = result.json
+    assert data["count"] == 2
+    assert data["status"] == "found"
+    prediction_data = data["predictions"]
+    assert prediction_data[1]["id"] == prediction2.id
```
alexgarel, line 420: Testing first prediction again is a good idea here:

```suggestion
# we still have first prediction
assert prediction_data[0]["id"] == prediction1.id
# but also the second
assert prediction_data[1]["id"] == prediction2.id
```

alexgarel (approved):
Ok let's merge it :-)

(I will just commit my suggestion first)

Issue #249, opened 2020-09-26
Detect AFDIAG products


teolemon:
https://static.openfoodfacts.org/images/products/370/056/645/3049/11.jpg
"FR-033-151"
The prefix is a company prefix. Trying to extract those could clash with regular packager codes. It could help that AFDIAG or gluten is likely to appear on the packaging.

https://docs.google.com/spreadsheets/d/1igQstTVR88ojwi82tcyCJm_I3LQOY59NKPjXcp1KviQ/edit#gid=223564579

teolemon:
Probably to be inserted here: https://github.com/openfoodfacts/robotoff/blob/master/robotoff/prediction/ocr/packager_code.py

teolemon:
FR 23.355.001 CE
FR XX.XXX.XXX CE FR XX.XXX.XXX EG FR XX.XXX.XXX EC
FR-XXX-XXX

PR #823, opened 2022-06-28
feat: Adding a regex for gluten packaging code

Jagrutiti:
### What
- Added a regex expression to test gluten code in packaging.

### Fixes
- #249

### DIFF
--- a/tests/unit/prediction/ocr/test_packager_code.py
+++ b/tests/unit/prediction/ocr/test_packager_code.py
@@ -10,7 +10,9 @@ from robotoff.prediction.ocr.packager_code import find_packager_codes
     [
         ("Sustainable palm oil RSPO-5068502 ", ["RSPO-5068502"]),
         ("RSPO-50685022", []),
-        ("QRSPO-2404885", []),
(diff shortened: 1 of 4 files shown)


alexgarel (changes requested):
We need to add test I think, because we are completely blind otherwise.

tests/unit/prediction/ocr/test_packager_code.py
```diff
@@ -11,6 +11,7 @@
         ("Sustainable palm oil RSPO-5068502 ", ["RSPO-5068502"]),
         ("RSPO-50685022", []),
         ("QRSPO-2404885", []),
+        ("FR-012-345", ["FR-012-345"]),
```
alexgarel, line 14: you should add some text where it does not match (eg FR-A22-322) and also put the code inside a sentence to verify it's working fine.

alexgarel: also fr-333-444 should not match (upper cased)

alexgarel (approved):

alexgarel:
:+1: @Jagrutiti

PR #832, opened 2022-07-21
feat: filter images and display in a list

Jagrutiti:
This PR is the second part of #618 

### What
- Collects all the image based on parameters passed. 

The detailed description of the PR is in the [comment here](https://github.com/openfoodfacts/robotoff/issues/618#issuecomment-1191170611). Pasting it here for easy reference:

> In image list (ImageCollection)
> I would like to have a filter, which is on by default, that just list images without ImagePrediction linked. that is:
> 
> without parameters, it would only return images without prediction images linked
> with parameter with_predicted=1 it would return all images
> The only other filter you can put there is the barcode

### Part of 
- #618


robotoff/app/api.py
```diff
@@ -1055,6 +1061,40 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
         resp.media = {"count": {"annotations": annotation_count}}
 
 
+def get_image_list_on_get(req: falcon.Request, resp: falcon.Response):
+    response: JSONType = {}
+    count: int = req.get_param_as_int("count", min_value=1, default=25)
+    page: int = req.get_param_as_int("page", min_value=1, default=1)
+    with_predicted: Optional[int] = req.get_param_as_int("with_predicted", default=1)
```
alexgarel, line 1068: default should be 0.

```suggestion
    with_predicted: Optional[int] = req.get_param_as_int("with_predicted", default=0)
```

robotoff/app/core.py
```diff
@@ -132,6 +132,34 @@ def get_insights(
...
+    query = ImageModel.select()
+
+    if where_clauses:
+        query = query.where(*where_clauses)
+
+    if with_predicted == 1:
+        # return all images
+        return query.iterator()
+
+    return query.iterator()
```
alexgarel, line 160: you forgot to take into account the count parameter.

robotoff/app/core.py
```diff
@@ -132,6 +132,34 @@ def get_insights(
...
+        where_clauses.append(ImageModel.barcode == barcode)
+
+    query = ImageModel.select()
+
+    if where_clauses:
+        query = query.where(*where_clauses)
+
+    if with_predicted == 1:
+        # return all images
+        return query.iterator()
```
alexgarel, line 158: with_predicted will be used to build the query so, no return here.

To build the query you should read https://docs.peewee-orm.com/en/latest/peewee/relationships.html

Hint: it's about outer join and using is_null() in where clause.
Do not hesitate to pdb in a test where you have created some Image and ImagePrediction, to try out the queries.

alexgarel (commented):
You are in the right direction :-)

robotoff/app/api.py
```diff
@@ -1055,6 +1061,69 @@ def on_get(self, req: falcon.Request, resp: falcon.Response, username: str):
...
+    response["count"] = get_images_(count=True)
+
+    if not images:
+        response["images"] = []
+        response["status"] = "no_images"
+    else:
+        response["images"] = images
+        response["status"] = "found"
+
+    resp.media = response
```
alexgarel, line 1092: you can delete this now, right :wink:

robotoff/app/core.py
```diff
@@ -132,6 +138,39 @@ def get_insights(
...
+    if barcode:
+        where_clauses.append(ImageModel.barcode == barcode)
+
+    query = ImageModel.select()
+
+    if with_predicted:
+        # return all images
+        query = query.join(ImagePrediction).where(ImagePrediction.image.is_null(True))
+    else:
+        query = query.join(ImagePrediction)
```
alexgarel, line 163: You don't need this in this case. Be it if we want to count ImagePrediction for each Image (which could be an idea in the future).

robotoff/app/core.py
```diff
@@ -132,6 +138,39 @@ def get_insights(
...
+    where_clauses = [ImageModel.server_domain == server_domain]
+
+    if barcode:
+        where_clauses.append(ImageModel.barcode == barcode)
+
+    query = ImageModel.select()
+
+    if with_predicted:
+        # return all images
+        query = query.join(ImagePrediction).where(ImagePrediction.image.is_null(True))
```
alexgarel, line 161: Kudos :muscle:

tests/integration/test_api.py
```diff
@@ -470,3 +477,45 @@ def test_annotate_insight_anonymous_then_authenticated(client, mocker):
         server_domain=settings.OFF_SERVER_DOMAIN,
         auth=OFFAuthentication(username="a", password="b"),
     )
+
+
+def test_get_images(client):
```
alexgarel, line 482: I forgot to do it in last PR but we should separate test of the API and test of the function.

Test on the function should be in test_core.py.

On the API you should try one empty test, but also a test with some data.

tests/integration/test_api.py
```diff
@@ -470,3 +477,45 @@ def test_annotate_insight_anonymous_then_authenticated(client, mocker):
...
+    result = client.simulate_get("/api/v1/images?page=1&count=25")
+    assert result.status_code == 200
+
+    image_model_factory1 = ImageModelFactory(barcode="123")
+    image_model_factory2 = ImageModelFactory(barcode="456")
+
+    image_prediction_factory1 = ImagePredictionFactory(image=image_model_factory1.id)
+    image_prediction_factory2 = ImagePredictionFactory(image=image_model_factory2.id)
```
alexgarel, line 490: For your information, thanks to factory we can also write it:

```suggestion
    image_prediction1 = ImagePredictionFactory(image__barcode="123")
    image_model1 = image_prediction1.image
    image_prediction2 = ImagePredictionFactory(image__barcode="456")
    image_model2 = image_prediction2.image
```

NB: you should not call the variables image_xxx_factory: the factory is the function that return an object (like a real life industrial factory). 
So you should prefer image_prediction1 and image_model1

tests/integration/test_api.py
```diff
@@ -470,3 +477,45 @@ def test_annotate_insight_anonymous_then_authenticated(client, mocker):
...
+    image_prediction_factory3 = ImagePredictionFactory()
```
alexgarel, line 491: For your test, you'd better want to have an image without ImagePrediction attached to it, to test your with_predicted=False

So here you'd better only create an image.
More over I will intentionally reuse barcode "123" to have two images returned on this barcode with with_predicted=True 

```suggestion
    image_model3 = ImageModelFactory(barcode="123")
```

robotoff/app/core.py
```diff
@@ -132,6 +138,39 @@ def get_insights(
     return query.iterator()
 
 
+def get_images(
+    with_predicted: Optional[bool] = False,
```
alexgarel, line 142: I think we should name this parameter (here and in api) `with_predictions` instead of `with_predicted`.

The idea is to return images that have no prediction. (not images that have not been predicted…)

tests/integration/test_api.py
```diff
@@ -470,3 +477,45 @@ def test_annotate_insight_anonymous_then_authenticated(client, mocker):
...
+
+    # test with "barcode" filter
+    image_model_data = get_images(barcode="123")
+    image_model_items = [item.to_dict() for item in image_model_data]
+    assert len(image_model_items) == 1
+    assert image_model_items[0]["id"] == image_model_factory1.id
+    assert image_model_items[0]["barcode"] == "123"
```
alexgarel, line 498: This test will work only if you create image_model3 without predictions as mentioned above.

And in thi case it will be image3.id that you get.

tests/integration/test_api.py
alexgarel: the image__barcode is a "trick" of Factory Boy, taken from django, which says set myobject.image.barcode to this value (note the double "_" between image and barcode)

tests/integration/test_api.py
```diff
@@ -470,3 +477,45 @@ def test_annotate_insight_anonymous_then_authenticated(client, mocker):
...
+
+    # test if "image_prediction" foreign key matches "image_model"
+    assert image_prediction_factory1.image.id == image_model_factory1.id
+    assert image_prediction_factory2.image.id == image_model_factory2.id
+
+    # test if foreign key does not match as not parameter is passed to image_prediction_factory3
+    assert not (image_prediction_factory3.image.id == image_model_factory1.id)
+    assert not (image_prediction_factory3.image.id == image_model_factory2.id)
```
alexgarel, line 506: you don't need this, you are not testing the factory !

tests/integration/test_api.py
```diff
@@ -470,3 +477,45 @@ def test_annotate_insight_anonymous_then_authenticated(client, mocker):
...
+
+    # test filter with "barcode" and "with_predicted=False"
+    image_prediction_data = get_images(barcode="123", with_predicted=False)
+    image_prediction_items = [item.to_dict() for item in image_prediction_data]
+    assert len(image_prediction_items) == 1
```
alexgarel, line 511: you already did this test above. instead test with `with_predicted=True` you should get two results this time.

tests/integration/test_api.py
```diff
@@ -470,3 +477,45 @@ def test_annotate_insight_anonymous_then_authenticated(client, mocker):
...
+
+    # test filter with "with_predicted=False"
+    image_prediction_data = get_images(with_predicted=False)
+    image_prediction_items = [item.to_dict() for item in image_prediction_data]
+    assert len(image_prediction_items) == 3
```
alexgarel, line 516: There again test with "with_predicted=True"

robotoff/app/core.py
```diff
@@ -132,6 +138,39 @@ def get_insights(
...
+
+    where_clauses = [ImageModel.server_domain == server_domain]
+
+    if barcode:
+        where_clauses.append(ImageModel.barcode == barcode)
+
+    query = ImageModel.select()
+
+    if with_predicted:
+        # return all images
```
alexgarel, line 160: It's the other way around (while the query below is ok)

```suggestion
    if not with_predicted:
        # return only images without prediction
```

alexgarel (commented):
You're on the right path.

alexgarel:
@Jagrutiti can you resolve conflicts and update ?

Then I will do a last commit to clean up some minor details (I'll explain to you after).

alexgarel (approved):

alexgarel:
@Jagrutiti we can merge when checks are done !

PR #834, opened 2022-07-29
feat: get image prediction list

Jagrutiti:
### What
This PR follows @alexgarel's [comment in #618](https://github.com/openfoodfacts/robotoff/issues/618#issuecomment-1198300543). Adding them below for easy reference:

>Let's move to ImagePredictionCollection !
>
>This is the same pattern as with ImageCollection.
>
>we have the barcode filter (but you will have to follow the "image" relation to get barcode)
>we also have the "type" filter
>This time instead of with_prediction, we have a with_logo (defaults to False), if False it discards all ImagePrediction for which  we have a logo annotation.
>Should be really the same id as for ImageCollection.

### Part of 
- #618


robotoff/app/core.py
```diff
@@ -165,6 +173,38 @@ def get_predictions(
...
+def get_image_predictions(
+    with_logo: Optional[bool] = False,
+    barcode: Optional[str] = None,
+    type: Optional[str] = None,
+    offset: Optional[int] = None,
+    count: bool = False,
+    limit: Optional[int] = 25,
+) -> Iterable[LogoAnnotation]:
+
+    query = LogoAnnotation.select().join(ImagePrediction).join(ImageModel)
```
alexgarel, line 185: You should not start from LogoAnnotation ? Because you want to return ImagePrediction, not LogoAnnotation !

robotoff/app/core.py
```diff
@@ -165,6 +173,38 @@ def get_predictions(
...
+
+    if barcode:
+        where_clauses.append(ImageModel.barcode == barcode)
+
+    if type:
+        where_clauses.append(LogoAnnotation.annotation_type == type)
+
+    if not with_logo:
+        # return only images without logo
+        query = query.where(ImagePrediction.image.is_null())
```
alexgarel, line 197: you have to use LefOuter join as for images but joining on LogoAnnotation this time.

tests/integration/test_api.py
```diff
@@ -569,3 +571,130 @@ def test_prediction_collection_no_filter(client):
...
+        "/api/v1/images/prediction/collection",
+        params={
+            "type": "label",
+            "with_logo": False,
+        },
+    )
+
+    assert result.status_code == 200
+    data = result.json
+    assert data["count"] == 0
```
alexgarel, line 700: If you use an ImagePredictionFactory, you will have a result.

The other tests above look ok to me.

robotoff/app/core.py
```diff
@@ -206,6 +207,42 @@ def get_predictions(
...
+    query = ImagePrediction.select()
+
+    where_clauses = []
+
+    if barcode:
+        query = query.join(ImageModel)
+        where_clauses.append(ImagePrediction.image.barcode == barcode)
+
+    if type:
+        where_clauses.append(LogoAnnotation.annotation_type == type)
```
alexgarel, line 229: we want to filter on ImagePrediction type, not  LogoAnnotation type.

robotoff/app/api.py
```diff
@@ -1141,6 +1142,38 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
...
 
 
+class ImagePredictionCollection:
+    def on_get(self, req: falcon.Request, resp: falcon.Response):
+        response: JSONType = {}
+        count: int = req.get_param_as_int("count", min_value=1, default=25)
+        page: int = req.get_param_as_int("page", min_value=1, default=1)
+        with_logo: Optional[bool] = req.get_param_as_bool("with_logo", default=False)
+        barcode: Optional[str] = req.get_param("barcode")
+        type: Optional[str] = req.get_param("type")
```
alexgarel, line 1152: We want to be compatible with parameters from InsightCollection, so we want a list of type.

```suggestion
        keep_types: Optional[List[str]] = req.get_param_as_list(
            "types", required=False
        )
```

This means changing get_image_predictions code

tests/integration/test_core_integration.py
```diff
@@ -19,6 +20,50 @@ def _set_up_and_tear_down(peewee_db):
     clean_db()
 
 
+def test_get_image_predictions():
+    logo_annotation1 = LogoAnnotationFactory(image_prediction__image__barcode="123")
+    logo_annotation2 = LogoAnnotationFactory(image_prediction__image__barcode="456")
+    logo_annotation = LogoAnnotationFactory(type="label")
```
alexgarel, line 26: As above if we want to be able to test all filters, we do want to use ImagePredictionFactory to have ImagePrediction instances without LogoAnnotation attached to them.

There again type that we care about is ImagePrediciton type, not LogoAnnotation type.

alexgarel (changes requested):
@Jagrutiti I made some comments hopping it will help you.

Issue #676, opened 2022-03-28
Have a route to list insight opportunities for a given type of tags


teolemon:
### What

The problem is that we have thousands of categories, so we should send thousands of requests to robotoff for just knowing which category has at least one question to answer.

A better option is to have a route returning something like that
```
{
	"remaining_category_to_validate": {
		  "en:waters": 10,
		  "fr:lasagnes": 3,
		  ...
	  }
}
```

Parameter would be type of insights (eg. type=category, or type=labels)

We should also be able to paginate, and sort on quantity.

If possible add an agribalyse=True filter (it use categories.json to get agribalyse and filter on them, eventually on python side).

### Why

Useful for games on bonus / malus for eco-score (avoid making 5 or 6 requests).

Useful to get data for eco-score (agribalyse)

### Part of
- #374

alexgarel:
This issue is solved.

PR #859, opened 2022-08-15
Filter insights opportunities based on type of tags

Jagrutiti:
### What
- This PR filters insight based on the number of questions unanswered filtered by tags.

### Fixes bug(s)
- #676

### DIFF
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -61,6 +61,8 @@ from robotoff.utils.i18n import TranslationStore
 from robotoff.utils.text import get_tag
 from robotoff.utils.types import JSONType
 from robotoff.workers.client import send_ipc_event
+from collections import Counter
+
 
 logger = get_logger()
 
@@ -1140,6 +1142,38 @@ class PredictionCollection:
 
         resp.media = response
 
+class UnansweredQuestionCollection:
+    def on_get(self, req: falcon.Request, resp: falcon.Response):
+        response: JSONType = {}
+        page: int = req.get_param_as_int("page", min_value=1, default=1)
+        count: int = req.get_param_as_int("count", min_value=1, default=25)
+        question_type: str = req.get_param("question_type")
+        
+        insights = list(
+            get_insights(
+                keep_types=[question_type],
+                annotation=0,
+                limit=count,
+            )
+        )
+
+
+        if not insights:
+            response["questions"] = []
+            response["status"] = "no_questions"
+        else:
+            response = Counter(insights[0].values())
+
+        resp.media = response
+
+
+
+
+
+
+        
+
+    
 
 cors = CORS(
     allow_all_origins=True,
@@ -1194,3 +1228,4 @@ api.add_route("/api/v1/dump", DumpResource())
 api.add_route("/api/v1/users/statistics/{username}", UserStatisticsResource())
 api.add_route("/api/v1/predictions/", PredictionCollection())
 api.add_route("/api/v1/images", ImageCollection())
+api.add_route("/api/v1/questions/unanswered/", UnansweredQuestionCollection())
--- a/tests/integration/models_utils.py
+++ b/tests/integration/models_utils.py
@@ -49,6 +49,7 @@ class ProductInsightFactory(UuidSequencer, PeeweeModelFactory):
     server_domain: str = factory.LazyFunction(lambda: settings.OFF_SERVER_DOMAIN)
     server_type = "off"
     unique_scans_n = 10
+    annotation = 0
 
 
 class PredictionFactory(PeeweeModelFactory):
--- a/tests/integration/test_api.py
+++ b/tests/integration/test_api.py
@@ -569,3 +569,40 @@ def test_prediction_collection_no_filter(client):
     assert prediction_data[1]["id"] == prediction2.id
     assert prediction_data[1]["type"] == "brand"
     assert prediction_data[1]["value_tag"] == "en:beers"
+
+
+def test_get_unanswered_questions_api(client):
+    result = client.simulate_get("/api/v1/questions/unanswered/")
+
+    assert result.status_code == 200
+    assert result.json == {"questions": [], "status": "no_questions"}
+
+    product1 = ProductInsightFactory(type="category", annotation=0)
+    result = client.simulate_get("/api/v1/questions/unanswered/", 
+        params={
+            "count": 5,
+            "page": 1,
+            "question_type": "category",
+        },
+    )
+    assert result.status_code == 200
+    data = result.json
+    assert len(data) == 2
+    assert len(data["questions"])==1
+    assert data["status"] == "found"
+    product_data = sorted(data["questions"], key=lambda d: d["id"])
+
+    product1 = ProductInsightFactory(type="label", annotation=0)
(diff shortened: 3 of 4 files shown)


tests/integration/test_api.py
```diff
@@ -569,3 +569,40 @@ def test_prediction_collection_no_filter(client):
...
+
+    assert result.status_code == 200
+    assert result.json == {"questions": [], "status": "no_questions"}
+
+    product1 = ProductInsightFactory(type="category", annotation=0)
+    result = client.simulate_get("/api/v1/questions/unanswered/", 
+        params={
+            "count": 5,
+            "page": 1,
+            "question_type": "category",
```
Jagrutiti, line 586: Hey @alexgarel 

I am not able to pass the parameters to the functions. `question_type` is always `None` when passed to `get_insights`

robotoff/app/api.py
```diff
@@ -1141,6 +1142,31 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
         resp.media = response
 
 
+class UnansweredQuestionCollection:
+    def on_get(self, req: falcon.Request, resp: falcon.Response):
+        response: JSONType = {}
+        page: int = req.get_param_as_int("page", min_value=1, default=1)
+        count: int = req.get_param_as_int("count", min_value=1, default=25)
+        question_type: str = req.get_param("question_type")
```
alexgarel, line 1150: we can name the parameter type to be consistent (although avoiding the name for the variable).

```suggestion
        question_type: str = req.get_param("type")
```

robotoff/app/api.py
```diff
@@ -1141,6 +1142,31 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
...
+
+        insights = list(
+            get_insights(
+                keep_types=[question_type],
+                group_by=True,
```
alexgarel, line 1155: I don't think it's a good option to have a boolean there. 

So either we put a string, the field we want to group on: here `group_by=Insight.value_tag` (which as the draw back of using Insight)

or we call the parameter group_by_value_tag (which is a bit less elegant but may better do the job here).

robotoff/app/core.py
```diff
@@ -121,6 +122,9 @@ def get_insights(
     if offset is not None and order_by != "random":
         query = query.offset(offset)
 
+    if group_by:
+        query = query.group_by(keep_types)
```
alexgarel, line 126: No you should see it's not ok because keep_types is a list of string while group_by expects a peewee field !

Here we want to group by values (value_tag precisely) that what the issue says.

So it's
```suggestion
        query = query.group_by(ProductInsight.value_tag)
```

But as soon as you do this you need to specify how you will aggregate the other fields (because they have different value), and really in our case we are just interested in counting elements. See [peewee documentation](https://docs.peewee-orm.com/en/latest/peewee/querying.html#aggregating-records) but in our case we do not group on one table versus a join table, but we really are interested only in two values: the value_tag and the count of entries, you can manage that using [right parameters to select and scalar as_tupe=True](https://docs.peewee-orm.com/en/latest/peewee/querying.html#retrieving-scalar-values)

The best to find the right query is to set a pdb.set_trace() in one of your test where you have some insight created and see what works.

tests/integration/models_utils.py
```diff
@@ -49,6 +49,7 @@ class Meta:
     server_domain: str = factory.LazyFunction(lambda: settings.OFF_SERVER_DOMAIN)
     server_type = "off"
     unique_scans_n = 10
+    annotation = 0
```
alexgarel, line 52: No annotation should be None by default… annotation = 0 means it has been annotated as "false". And we don't want such a default.

If you need it locally, use it as a parameter `ProductInsightFactory(annotation=0)`

tests/integration/test_api.py
```diff
@@ -569,3 +569,36 @@ def test_prediction_collection_no_filter(client):
...
     assert prediction_data[1]["value_tag"] == "en:beers"
+
+
+def test_get_unanswered_questions_api(client):
+    result = client.simulate_get("/api/v1/questions/unanswered/")
+
+    assert result.status_code == 200
+    assert result.json == {"questions": [], "status": "no_questions"}
+
+    product1 = ProductInsightFactory(type="category", annotation=0)
```
alexgarel, line 580: here to test the API the best is to create at least 3 or 5 examples with 2 or 3 different value_tag properties.

robotoff/app/api.py
```diff
@@ -1141,6 +1142,31 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
...
+                annotation=0,
```
alexgarel, line 1156: No what we really want are unanswered questions, that is with annotation=None.

See the documentation of this attribute in models.py:

```python

    # The annotation of the given insight. Four possible values are possible:
    # null = This insight has not been annotated
    #  -1 = Rejected
    # 0 = 'I don't know'
    # 1 = Validated
    annotation = peewee.IntegerField(null=True, index=True)
```

tests/integration/test_api.py
alexgarel: It's working, but I think you pdb in the first `simulate_get` above at line 576, where this parameter is None ! (that's why this is a good idea to separate tests ;-) )

tests/integration/test_api.py
```diff
@@ -569,3 +569,36 @@ def test_prediction_collection_no_filter(client):
     assert prediction_data[1]["id"] == prediction2.id
     assert prediction_data[1]["type"] == "brand"
     assert prediction_data[1]["value_tag"] == "en:beers"
+
+
+def test_get_unanswered_questions_api(client):
+    result = client.simulate_get("/api/v1/questions/unanswered/")
+
+    assert result.status_code == 200
+    assert result.json == {"questions": [], "status": "no_questions"}
```
alexgarel, line 578: As for other PR, you should separate this first test from the second.

```suggestion
def test_get_unanswered_questions_api_empty(client):
    result = client.simulate_get("/api/v1/questions/unanswered/")

    assert result.status_code == 200
    assert result.json == {"questions": [], "status": "no_questions"}

def test_get_unanswered_questions_api(client):
```

tests/integration/test_core_integration.py
```diff
@@ -114,3 +115,37 @@ def test_get_images():
...
+
+    product4 = ProductInsightFactory(type="category", value_tag="en:chicken")
+    insight_data4 = get_insights(keep_types=["category"], value_tag="en:chicken")
+    insight_data_items4 = [item.to_dict() for item in insight_data4]
+    assert insight_data_items4[0]["id"] == product4.id
+    # assert insight_data_items4[0]["type"] == []
+
+    insight_data5 = get_insights(keep_types=["category"])
+    insight_data_items5 = [item.to_dict() for item in insight_data5]
+    assert len(insight_data_items5) == 2
```
alexgarel, line 147: You should first create all your data, and then do one or two call to get the lists.

you need insight spread on:
- at least two types
- different value_tag (but with repeated values)
- different barcodes

alexgarel (commented):
@Jagrutiti some comments to put you back on track.

alexgarel (approved):

alexgarel:
@Jagrutiti thanks, you now only have to merge master before we can merge it back !

PR #882, opened 2022-08-31
feat: Get Logo Annotation list 

Jagrutiti:
### What
This PR does the following [from here](https://github.com/openfoodfacts/robotoff/issues/618#issuecomment-1231879826): 

> Next is LogoAnnotationCollection
>
> It's the same thing, with as previous view but for LogoAnnotation:
filters on:
>
> 
> 
> - barcode
> - type filter (a list of type) keep_types (filter on annotation_type)
> - value_tag (filter on annotation_value_tag)

### Part of 
- #618

### DIFF
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -25,6 +25,7 @@ from robotoff.app.core import (
     get_image_predictions,
     get_images,
     get_insights,
+    get_logo_annotation,
     get_predictions,
     save_annotation,
 )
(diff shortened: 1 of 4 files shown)


robotoff/app/api.py
```diff
@@ -1177,6 +1178,41 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
         resp.media = response
 
 
+class LogoAnnotationCollection:
+    def on_get(self, req: falcon.Request, resp: falcon.Response):
+        response: JSONType = {}
+        barcode: Optional[str] = req.get_param("barcode")
+        keep_types: Optional[List[str]] = req.get_param_as_list("types", required=False)
```
Jagrutiti, line 1185: I am naming the parameter as `types`. But in the test case it only works when I pass it as `keep_types`

tests/integration/test_api.py
```diff
@@ -651,3 +651,101 @@ def test_image_prediction_collection(client):
...
+        )
+
+    result = client.simulate_get(
+        "/api/v1/annotation/collection?count=5&page=1&keep_types=label"
+    )
+
+    data = result.json
+    assert data["count"] == 12
+    assert data["status"] == "found"
+    assert [q["annotation_value_tag"] for q in data["annotation"]] == [
```
Jagrutiti, line 745: This test case fails. I should get 5 records but I am getting 12 records all the time.

PR #887, opened 2022-09-03
fix: Added server_domain all "Collection" class

Jagrutiti:
### What
- Added `server_domain` to all "Collection" class ([First step from here](https://github.com/openfoodfacts/robotoff/issues/618#issuecomment-1235475222))

### Part of 
- #618


alexgarel (approved):
LGTM.

We should be able to remove the sorting when the #888 will be merged.

PR #888, opened 2022-09-03
feat: Sort collection API

Jagrutiti:
### What
- The instruction of the PR is [here](https://github.com/openfoodfacts/robotoff/issues/618#issuecomment-1235475222): 
>by default, always sort on freshest item (higher id ?) first

### Part of 
- #618

### DIFF
--- a/robotoff/app/core.py
+++ b/robotoff/app/core.py
@@ -134,6 +134,9 @@ def get_insights(
 
         elif order_by == "n_votes":
             query = query.order_by(ProductInsight.n_votes.desc())
+    else:
+        query = query.order_by(ProductInsight.timestamp.desc())
+
     if as_dict:
         query = query.dicts()
 
@@ -201,6 +204,8 @@ def get_predictions(
     if where_clauses:
         query = query.where(*where_clauses)
 
+    query = query.order_by(Prediction.timestamp.desc())
+
     if count:
         return query.count()
     else:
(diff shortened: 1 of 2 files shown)


robotoff/app/core.py
```diff
@@ -134,6 +134,9 @@ def get_insights(
 
         elif order_by == "n_votes":
             query = query.order_by(ProductInsight.n_votes.desc())
+    else:
+        query = query.order_by(ProductInsight.id.desc())
```
alexgarel, line 138: As said in meeting: better not introduce a default sorting in a widely used function, it might have unexpected side effects.

We should instead use `order_by="most_recent"` in Collection view and implement it here.

Also you should use timestamp here and not id, because id is special on ProductInsight (it's not numerical, it's a uuid).

robotoff/app/core.py
```diff
@@ -363,6 +369,8 @@ def get_logo_annotation(
     if offset is not None:
         query = query.offset(offset)
 
+    # query = query.order_by(LogoAnnotation.image_prediction.timestamp.desc())
```
alexgarel, line 372: here you can use LogoAnnotation.id for order_by. `id` exists because it's automatically added by peewee.

See https://docs.peewee-orm.com/en/latest/peewee/models.html#fields
> In the above example, because none of the fields are initialized with primary_key=True, an auto-incrementing primary key will automatically be created and named “id”. Peewee uses [AutoField](https://docs.peewee-orm.com/en/latest/peewee/api.html#AutoField) to signify an auto-incrementing integer primary key, which implies primary_key=True.

alexgarel (approved):
Kudos. 

You need to resolve conflicts and we should be able to merge.

PR #862, opened 2022-08-29
feat: send reported images to alexandre server

alexfauquette:
Not sure it works, I did not setup properly the project for now, and it's crappy

PR #889, opened 2022-09-07
feat: Add image moderation service

alexgarel:
Send flagged images to an external moderation database.

Replaces #862

### DIFF
--- a/robotoff/slack.py
+++ b/robotoff/slack.py
@@ -19,11 +19,11 @@ class SlackException(Exception):
     pass
 
 
-class SlackNotifierInterface:
-    """SlackNotifierInterface is an interface for posting Robotoff-related alerts and notifications to the OFF Slack channels."""
+class NotifierInterface:
+    """NotifierInterface is an interface for posting Robotoff-related alerts and notifications to the OFF Slack channels."""
 
     def notify_image_flag(
-        self, predictions: List[Prediction], source: str, barcode: str
+        self, predictions: List[Prediction], source_image: str, barcode: str
     ):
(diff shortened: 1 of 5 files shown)


alexgarel:
@alexfauquette I recoded your PR in a clean way, because I think we will move on to that however. (and I prefer to add some tests before going live).

On question though: in slack notifier we do not send notification if there is no prediction linked with the image. Do we want the same here or should we always send it ?

alexfauquette:
I assume if we send images to the server event when there is no prediction we will end up sending all the new pictures to the server which is not the desired behavior

robotoff/slack.py
```diff
@@ -96,7 +106,57 @@ def _slack_message_block(
...
+    ):
+        """Send image to the moderation server so that a human can moderate it"""
+        image_url = settings.OFF_IMAGE_BASE_URL + source_image
+        image_id = int(source_image.rsplit("/", 1)[-1].split(".", 1)[0])
+        params = {"imgid": image_id, "url": image_url}
+        http_session.put(f"{self.service_url}/{barcode}", data=params)
+
+
+class SlackNotifier(NotifierInterface):
     """SlackNotifier implements the real SlackNotifier."""
```
Jagrutiti, line 168: Since you are implementing NotifierInterface.

Shouldn't the comment be: ` """SlackNotifier implements the real NotifierInterface."""`

Rest looks good to me.

Jagrutiti (changes requested):

Jagrutiti (approved):

Issue #376, opened 2021-08-23
Support the Eco-Score through increased data extraction


teolemon:
#### Origins
- [ ] #306 
#### Recycling
- [ ] #260
- [ ] #257 
### Why

- The Eco-Score is very data-intensive.

### Part of
#374

Issue #306, opened 2020-12-31
Extract origins for the Eco-Score


teolemon:
Géographical text extraction to insights, that if validated, go to the `origins` field. This will help get origins for the eco-score computation.

Questions: "Is the following statement valid ?"

“Importé du {Pérou} >> en:Ingredients from Peru
“{Quinoa} origine non-UE” (séparé de la liste des ingrédients) >> en:Quinoa from outside Europe
“Certains ingrédients ne proviennent pas de {France}” >>en:Non {{French}} ingredients
“à partir de blé dur de diverses origines” >> en:Wheat from various origins

senthilnathan-ra:
@teolemon I'm new to the community and want to contribute. Is this issue still open?

Jagrutiti:
Hey @teolemon

How do I get the geographical information? Is there an API for it? Could you please shed more light on this? 

Thanks :)

PR #890, opened 2022-09-08
feat: Added origin extraction

Pykorm:
feat: Add ocr origin extraction
### What
- This pull request adds the origin extraction from ocr, for the eco-score. It also adds an api endpoint, to extract origin from any ocr url. It uses one regex by lang (currently French and partially english are support), and this regex is made to match almost any sentence that tells something about the origin of any ingredient. See #306.

- Explanations of the french regex, same reasoning in english :
In french, a sentence always starts with the subject, so we first start with "({INGREDIENTS_FR_JOINED})" that matches any ingredient.
But we can have several ingredients, with 'and' or a coma as separator, like in a sentence like "Caramel and chocolate are made in France". That's why we add  "?(?:et |, ?)?" that means "it may be a space, then a 'and' or a coma, but maybe there is nothing"
In a sentence like "Made in France", there is no subject, so all this is optionnal. It can also be repeated like in "Caramel, Chocolate and Wheat are from France". That's why we use a "*" quantifier 
Then there is a support for sentence like "100% made in France" or "quinoa is 100% French" with the ?(?:100%)?
Then the verb, that is also optional, like in "french quinoa". In sentence like "made with french quinoa", the verb "made" isn't captured by the regex, but "french quinoa" is.
After that we have some connectors like "dans, depuis" (= in, from), and also their negations, "hors, en dehors" (=outside). Those are also optional, but the regex can match even if they are in the middle of the sentence.
They are followed by other connectors like "de la|de l' |de|du" (="of" adapted for the different genders)
Then we have the country as a name "?(?P<country>{COUNTRIES_FR_JOINED})?", like in "made in France". 
and after we have the country as a adjective , like in "french quinoa" (in French, the adjective goes after the noun). "?(?P<country_adj>{COUNTRIES_ADJECTIVES_FR_JOINED})?"
and after we have "?(?P<large_origin>(?:divers(?:es)?|diff[ée]rent(?:es)?|d'autres) ?(?:pays|[ée]tats?|r[ée]gions?|continents?|origines?))?" that supports sentence like "quinoa from different countries".
Finally, you may have noticed that everything is optional. That means that this regex match empty strings. To avoid that we add "\b(?<!\w)" at the beginning and "\b(?!\w)" at the end. Idea comes from https://stackoverflow.com/a/50188154/8966453 

- Here are some sentence that matches :
"quinoa de France"
"quinoa produit en France"
"quinoa cultivée en dehors de l' Union Européenne"
"Fabriqué en France"
"Tout les ingrédients sont fabriqués en France"
"A partir de quinoa français"
"quinoa 100% français"
"Le quinoa a différentes origines"
"Le quinoa provient de different pays"
 - NOT SUPPORTED:
"Le quinoa de notre recette provient de different pays" ("de notre recette" is a genetive, hard to support)
"Nos agriculteurs français produisent le blé de notre recette" (subject isn't the ingredient, hard to support)

- I added the file [data/ocr/countries.json](https://github.com/Pykorm/robotoff/blob/master/data/ocr/countries.json) which is a slightly modified version of the file at https://static.openfoodfacts.net/data/taxonomies/countries.full.json. I added a "nationalities" field for every countrie, which contains the name we give to the inhabitants, key is the lang and value is the nationality name. There is currently only a few countries in english and french. This is mandatory to support sentences like "french quinoa". My code also uses the[ categories.full.json](https://github.com/openfoodfacts/robotoff/blob/master/data/taxonomies/categories.full.json) file.

### Note :
- This is my first contribution to an open source project, I tried to do things well. Every tests passes. I started from the code in [nutrient.py ](https://github.com/openfoodfacts/robotoff/blob/master/robotoff/prediction/ocr/nutrient.py) as suggested in #306. I added some constants to settings.py, an api endpoint, and a file in data. 

### Part of 
- #376  
- #600


robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,264 @@
...
+                if ing_ori["origin"] == origin:
+                    origin_index = index
+            
+            if origin_index == -1:
+                ingredients_origins.append({
+                    "origin": origin,
+                    "same_for_all_ingredients": True, # True unless group "ingredients" matched
+                    "concerned_ingredients": None
+                })
+                origin_index = len(ingredients_origins) - 1
```
alexgarel, line 178: why not use a dict (that at the end you will transform to a list of prediction) so that you can index by origin, instead to have to scan the list ? It would make the code more readable.

You could even use a collections.defaultdict:
```python
ingredients_origins: Dict[str, Dict[str, Any]] = collections.defaultdict(
    lambda: {"same_for_all_ingredients": True, "concerned_ingredients": None}
)
```

and here:
```python
origin_data = ingredients_origins[origin]
```

At the end you would use an iterator on items()

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,264 @@
...
+
+def check_if_general_word_in_ingredients(general_words_regex: List[str], ingredients: str):
+    """
+    Utility function for checking if one of the general words like 'elements' or 'ingredients' is 
+    present in the list of ingredients returned by extract_ingredients_from_match_string
+    """
+    for reg in general_words_regex:
+        if re.search(reg, ingredients) is not None:
+            return True
+    return False
```
alexgarel, line 212:
```suggestion
    return any(re.search(reg, ingredients) is not None for reg in general_words_regex)
```

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,264 @@
...
+def extract_ingredients_from_match_string (ingredients_text: str, lang) -> List[str]:
+    """
+    Utility function that splits the regex's matched group "ingredient" into a list of ingredients.
+    the regex matches with sentence with several ingredients, like 'quinoa and rice are made in France'
+    It match the whole "quinoa and rice " string, and this function splits each ingredients.
+    """
+    splitted = ingredients_text.split(" ")
+    def is_not_empty (s):
+        return s != ""
+    splitted = filter(is_not_empty, splitted)
```
alexgarel, line 249: Shouldn't you simply use split() as it is more powerful (split on multiple space and any type of spaces).
```
>>> "a   b  c \r\t   d \n\t".split()
['a', 'b', 'c', 'd']
```

```suggestion
    splitted = ingredients_text.split()
```

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,264 @@
...
+        return s[0].upper() + s[1:].lower()
+    standardized = map(standardize_ingredient, splitted) 
+    def to_id (s):
+        try:
+            return next(ingredient_id for ingredient_id, ingredient in INGREDIENTS.items() 
+                if lang in ingredient["name"]
+                and ingredient["name"][lang] == s
+            )
+        except StopIteration:
+            return s
```
alexgarel, line 260: As we know we will do this kind of lookup, I would build a dict at initialization time, so that we can just check the key directly, it will really improve performance.
So that I can do 
```python
return INGREDIENTS_ID[lang].get(s, s)
```

building the INGREDIENTS_ID is something like:
```python
INGREDIENTS_ID = collections.defaultdict(dict)
for ingredient_id, ingredient in INGREDIENTS.items():
    for lang, name in ingredient["name"].items():
        INGREDIENTS_ID[lang][name] = ingredient_id
````

File: tests/unit/prediction/ocr/test_origin.py [hunk 5]
```diff
@@ -0,0 +1,85 @@
...
+        eget mollis nisi felis id ipsum. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae""", 
+        [{"origin": "en:france", "same_for_all_ingredients": False, "concerned_ingredients": ["en:quinoa", "en:rice"]},
+         {"origin": "en:germany", "same_for_all_ingredients": False, "concerned_ingredients": ["en:wheat"]}])
+    ],
+)
+def test_find_origin(text: str, origins: List[Dict[str, Union[str, List[str], bool]]]):
+    results = find_origin(text)
+    for index, insight in enumerate(results):
+        assert "ingredients_origins" in insight.data
+        assert insight.data["ingredients_origins"] == origins[index]
```
alexgarel, line 85: Really cool to have added all those tests !

alexgarel (changes requested):
@Pykorm first of all a big big thanks for this courageous PR that will really help improve on the eco-score front !

I added quite structural comments, because it's always easier reading code to see improvements, but I hope it won't discourage you.

I think we can concentrate on having the Predictions in this PR, and deploy it asap to get predictions.

That said I wanted to make you aware, as this is a new type of prediction that we will then have to:
- add an importer to generate insights
- add an annotator to apply validated insights on open food facts.

But this is better done in a separate PR.

alexgarel:
I forgot to add two things @Pykorm:

1. you should use make lint and make checks to have your code in good shape :-)
2. it's ok to have the json of countries in the code for now, but it would be cleaner if:
    1. we add needed information in the openfoodfacts taxonomy for countries as properties
    2. we use the taxonomy json through the Taxonomy class in robotoff
    
    this insures having future updates of the taxonomy

robotoff/app/api.py
```diff
@@ -365,6 +365,40 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
...
+            raise falcon.HTTPBadRequest("a JSON file is expected")
+
+        barcode = get_barcode_from_url(ocr_url)
+
+        if barcode is None:
+            raise falcon.HTTPBadRequest(f"invalid OCR URL: {ocr_url}")
+
+        try:
+            predictions = extract_ocr_predictions(
+                barcode, ocr_url, [PredictionType.origin]
```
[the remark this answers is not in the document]
Pykorm: Do you mean I should replace 
/api/v1/predict/origin
/api/v1/predict/nutrient
/api/v1/predict/ocr_insights

by /api/v1/predict/ that takes a "type" parameter (that represent the requested predictions's types and may appear more than once), and returns the list of predictions ? I just copy and adapt the code of the NutrientPredictorResource without thinking to much about it, but maybe it's a good idea.

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,264 @@
...
+#     - inside the specified country : "dans" "depuis" "en" 
+#     - outside the specified country : "hors" "en dehors"
+#     if it is None, you can safely think of it as "inside the specified country"
+# if none of the three groups below is found in the regex, then ignore the match, it does'nt tell anything about the origin
+# - "country" : (Optional) the place from which the product comes as a noun
+# - "country_adj" : (Optional) the place from which the product comes as an adjective
+# - "large_origin" : (Optional) (boolean) the place from which the product comes is large, or its ingredients come from several different countries
+FR_ORIGIN_REGEX = rf"\b(?<!\w)(?P<ingredients>(?:(?:{INGREDIENTS_FR_JOINED}) ?(?:et |, ?)? ?)*)? ?(?:100%)? ?(?P<negation>n'|ne) ?(?:est|sont|a|ont)? ?(?:{VERBS_FR_JOINED})? ?(?:pas|plus)? ?(?P<in_or_outside>dans|depuis|hors|en dehors|en)? ?(?:de fabrication|d'|de la|de l' |de|du)? ?(?:(?P<country>{COUNTRIES_FR_JOINED})|(?P<country_adj>{COUNTRIES_ADJECTIVES_FR_JOINED})|(?P<large_origin>(?:divers(?:es)?|diff[ée]rent(?:es)?|autres|un autre|plusieurs autres|plusieurs) (?:pays|[ée]tats?|r[ée]gions?|continents?|origines?)))\b(?!\w)"
```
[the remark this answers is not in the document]
Pykorm: No I did'nt know verbose flags ! Thank you, I will update those comments, it will be much better.

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,264 @@
...
+GENERAL_WORDS_FR = ["ingredients?", "[ée]lements?", "composition", "production", "mati[èe]res? premi[èe]res?"]
+INGREDIENTS_FR.extend(["ingr[ée]dients?", "[ée]l[ée]ments?", "composition", "production", "mati[èe]res? premi[èe]res?"])
+INGREDIENTS_FR_JOINED = "|".join(INGREDIENTS_FR)
+
+COUNTRIES_IN_ALL_LANGS = {}
+with open(settings.OCR_COUNTRIES_IN_ALL_LANGS, "r") as file:
+   COUNTRIES_IN_ALL_LANGS = json.loads(file.read())
+
+COUNTRIES_FR = [coutry["name"]["fr"]
+for coutry in COUNTRIES_IN_ALL_LANGS.values()
```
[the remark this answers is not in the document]
Pykorm: Oups :/

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,264 @@
...
+OUTSIDE_WORDS_BY_LANG = {
+    "fr": OUTSIDE_WORDS_FR,
+    "en": OUTSIDE_WORDS_EN
+}
+
+LARGE_ORIGIN = "large origin" # large origin means several countries, or an unknown country outside of the local area. 
+# (ex: "quinoa comes from outside the E.U", "quinoa has several origins")
+UNKNOW_ORIGIN = "unknow origin"
+ALL_INGREDIENTS = "all ingredients"
+
```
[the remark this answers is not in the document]
Pykorm: Yep, good idea, I'll be working on it

robotoff/prediction/ocr/origin.py
Pykorm: Yes it's clearly what should be done, I don't why I was scanning the list each time

robotoff/prediction/ocr/origin.py
Pykorm: Good idea, I'll do this while making all this a class

robotoff/app/api.py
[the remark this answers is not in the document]
alexgarel: hum, I didn't so the other one.

Ok for now let's keep it like it, and we will adapt on a separate request.

robotoff/prediction/ocr/origin.py
alexgarel: It's always easier to see this kind of thing as a reviewer !

Pykorm:
Just commit the modifications you suggested. Now we have a OriginParser class, with lazy initialization.

How can I use the taxonomy properly ? To create a store in [taxonomy.py](https://github.com/openfoodfacts/robotoff/blob/master/robotoff/taxonomy.py) I need a url (and a fallback path). For example, the label taxonomy store points to http://static.openfoodfact.[domain]/labels.full.json
So if I want to implement a new store, for countries, the file must exists at some url, isn't it ? 

[countries.json](https://github.com/Pykorm/robotoff/blob/master/data/ocr/countries.json) is a modified version of the file at https://static.openfoodfacts.net/data/taxonomies/countries.full.json. Basicly it's the same file with a "nationalities" field added for every country, that may contains the nationality of the inhabitants of this countries, by lang. It would be great to add these modification to the the file at https://static.openfoodfacts.net/data/taxonomies/countries.full.json, but as far as I understand, it's not robotoff that serves it. So how can I make use of the openfoodfacts taxonomies, and make this code cleaner ?

stephanegigandet:
Hi @Pykorm , that's very interesting, thank you for your contribution!

In Product Opener (the OFF backend written in Perl, also with a lot of regular expressions), we also have some functions to extract things like "Origine du Cacao : Pérou", but the match is not as complete as the one you made. I'm not using the ingredients and country taxonomies, so it's harder to see some patterns with multiple ingredients and/or multiple countries.

The corresponding code is there: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Ingredients.pm#L1003

stephanegigandet:
Regarding 

```
nationalities : {
en: "Hungarian",
fr: "hongrois"
}
```

We could add something like that in the OFF taxonomies, as properties. We have to think a bit about it, as we may want to cover the nationalities names for different gender and number (e.g. "grec", "grecque", "grecs", "grecques") so that we match "olives grecques".

One thing I'm wondering about is performance of a regex with 10k ingredients (and synonyms) + 200 nationalities. I've never used regexps that big, but maybe that's not an issue at all.

Pykorm:
Hello, 
You're right, we should care about gender and plural. Maybe we could store the nationalities as regex, something like that:
`{
  "fr": "grec(?:que)?s?"
}`
but it would be hard to create or find a database with all those names. 

I'm not woring about regexes performance, as there is no better solution. I don't see any way to recognize the sentences that say something about the origin that doesn't imply looping over every verbs, every country name... Those computations are mandatory. And I think that the performance of the python's re module are the best we could have.

temp.txt
```diff
@@ -0,0 +1,339 @@
+Angleterre
```
raphael0202, line 1: Is it expected that these two files temp.txt and temp2.txt are included?

Makefile
```diff
@@ -2,7 +2,7 @@
 
 # nice way to have our .env in environment for use in makefile
 # see https://lithic.tech/blog/2020-05/makefile-dot-env
-# Note: this will mask environment variable as opposed to docker-compose priority
+# Note: this will mask environment variable as opposed to docker compose priority
```
raphael0202, line 5: why have you introducted these changes?

raphael0202:
Can you also move `countries.json` to `data/taxonomies`? `data/ocr` is reserved for pattern/blacklist files used in OCR detections.

.env
```diff
@@ -8,8 +8,8 @@ PO_LOCAL_NET=po_default
 
 # User uid - in dev align this with your own user uid / gid
 # see https://gist.github.com/alexgarel/6e6158ee869d6db2192e0441fd58576e
-# OFF_UID=1000
-# OFF_GID=1000
+ OFF_UID=1000
+ OFF_GID=1000
```
raphael0202, line 12: why have you introduced these changes?

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,374 @@
...
+    def __init__(self) -> None:
+        self.initialized = False
+
+    def initialize(self):
+        self.INGREDIENTS = {}
+        INGREDIENTS = json.load(open(settings.TAXONOMY_CATEGORY_PATH, "r"))
+
+        # French ----------------
+        INGREDIENTS_SYNONIMS_FR = [
```
raphael0202, line 30: INGREDIENTS_SYNONIMS_FR -> INGREDIENTS_SYNONYMS_FR

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,374 @@
...
+                )
+                if ingredients is not None:
+                    ingredients_origins[origin]["same_for_all_ingredients"] = False
+                    ingredients_origins[origin]["concerned_ingredients"] = []
+                    for ingredient in ingredients:
+                        ingredients_origins[origin]["concerned_ingredients"].append(
+                            ingredient
+                        )
+
+        if len(ingredients_origins) == 0:
```
raphael0202, line 287: These lines are not necessary

robotoff/prediction/ocr/origin.py
```diff
@@ -0,0 +1,374 @@
...
+        |(?P<large_origin>(?:divers(?:es)?|diff[ée]rent(?:es)?|autres|un\sautre|plusieurs\sautres|plusieurs)\s(?:pays|[ée]tats?|r[ée]gions?|continents?|origines?)))\b(?!\w)'''
+        FR_ORIGIN_OCR_REGEX = OCRRegex(
+            re.compile(FR_ORIGIN_REGEX, flags=re.IGNORECASE|re.VERBOSE),
+            field=OCRField.full_text_contiguous,
+            lowercase=True,
+        )
+
+        # English -----------------------
+
+        INGREDIENTS_SYNONIMS_EN = [
```
raphael0202, line 150: INGREDIENTS_SYNONIMS_EN -> INGREDIENTS_SYNONYMS_EN

temp.txt
Pykorm: Oh no ! I just missclicked in vscode and push everything, I was thinking that I managed to clean my mistake, but apparently not :/

Makefile
Pykorm: Same thing, I made that because on my computer for unknown reason the "docker compose" don't exists, I have to use "docker-compose". This should be removed.

robotoff/prediction/ocr/origin.py
Pykorm: You're right, it was usefull in old commits and is now useless

Pykorm:
I just commited the changes you suggested, thank you for your review.

.env
```diff
@@ -56,4 +56,4 @@ SENTRY_DSN=
 IPC_AUTHKEY=ipc
 IPC_HOST=0.0.0.0
 IPC_PORT=6650
-WORKER_COUNT=8
+WORKER_COUNT=8
```
raphael0202, line 59: Can you revert to the original version for this file too?

Pykorm: Yep !

raphael0202:
Sorry for the (very) late reply! Can you give us an estimate of how much time does it take to process one OCR JSON?
It's important to know this before merging in production :)

Pykorm:
Is it possible to update the doc in this PR ? As it was created before de file api.yml, I can't edit this file. If I create it at the same path, would it be merge correctly ?

One OCR processing take about 0.00442 s, or 4.442ms. It was run on my laptop, with a 
i7-1260P (12th gen) and 16go of RAM.

raphael0202:
There are some conflicts with master, you need to merge openfoodfacts/robotoff master branch into your branch first.

> One OCR processing takes about 0.00442 s, or 4.442ms.
Perfect :)

Pykorm:
Sorry, i accidently closed the PR :/

raphael0202:
There are stillmany conflicts (and unwanted changes introduced by the merge). I think it has to do with the fact you're working on master, and not on a feature branch.
I've created a new branch with all your modifications here: https://github.com/openfoodfacts/robotoff/tree/origin-extraction
Is it ok if I merge this one instead?

Pykorm:
Yes, thank you a lot !

Issue #78, opened 2019-08-06
Extract AOC from images


teolemon:
### What
- Extract AOC from images to match AOC categories (especially for wine bottles)
  - "Appellation Clairette de Dié contrôlée"
  - "Appellation * contrôlée"
  - https://static.openfoodfacts.org/images/products/325/622/711/0559/1.jpg
  - https://static.openfoodfacts.org/images/products/325/622/711/0559/1.json
- Save as a category insight
### Part of
- https://github.com/openfoodfacts/robotoff/issues/600

alexgarel:
see also labels taxonomy to see the translation in different languages

raphael0202:
Implemented, closing the issue.

PR #891, opened 2022-09-09
feat: Extraction of category by REGEX on AOP

GabrielBeFr:
### What
- To answer the issue #78 , I added created a file category_from_AOC.py in robotoff/prediction/ocr/ to use REGEX in order to get a sure prediction of the category of products with an AOC syntax such as "Appellation Clairette de Die Protégée".
- I added an import in the file robotoff/prediction/ocr/core.py to take it into account.
- I added a unit test in tests/unit/prediction/ocr named test_category_from_AOC.py to test the former file.

- For now, since the REGEX have not been deeply developped, it will only match things like "Appellation Bordeaux Contrôlée/Protégée" or "Camembert Appellation d'origine protégée". I'm currently adding new REGEX for international syntaxes.
- However, an issue that I encountered without being able to fix it is if multiple spaces occur in the ocr result. For example, it will not match "Appellation Clairette    de Die contrôlée". I thought the field OCRField.full_text_contiguous would be able to correct it but it doesn't seem so.  

### Part of 
- #78

### DIFF
--- a/robotoff/off.py
+++ b/robotoff/off.py
@@ -500,3 +500,25 @@ def select_rotate_image(
 
     r.raise_for_status()
     return r
+
+
+def normalizing(line, lowercase=True):
+    # removing accents
+    line = re.sub(r"[¢£¤¥§©ª®°²³µ¶¹º¼½¾×‰€™]", '-', line)
+    line = re.sub(r"[éè]", 'e', line)
+    line = re.sub(r"[à]", 'a', line)
+    line = re.sub(r"[ù]", 'u', line)
+    # changing unwanted character to "-"
+    line = re.sub(r"&\w+;", "-", line)
+    line = re.sub(
+                r"[\s!\"#\$%&'()*+,\/:;<=>?@\[\\\]^_`{\|}~¡¢£¤¥¦§¨©ª«¬®¯°±²³´µ¶·¸¹º»¼½¾¿×ˆ˜–—‘’‚“”„†‡•…‰‹›€™\t]",  # noqa: E501
+                "-",
+                line,
+            )
+    # lowering the line if wanted
+    if lowercase:
+        line = line.lower()
+    # removing excess "-"
+    line = re.sub(r"-+", "-", line)
+    line = line.strip("-")
+    return line
--- a/robotoff/off.py
+++ b/robotoff/off.py
@@ -502,23 +502,27 @@ def select_rotate_image(
     return r
 
 
-def normalizing(line, lowercase=True):
+def normalize_tag(value, lowercase=True):
+    """given a value normalize it to a tag (as in taxonomies)
+
+    This means removing accents, lowercasing, replacing spaces with dashes, etc..
+    """
     # removing accents
-    line = re.sub(r"[¢£¤¥§©ª®°²³µ¶¹º¼½¾×‰€™]", "-", line)
-    line = re.sub(r"[éè]", "e", line)
-    line = re.sub(r"[à]", "a", line)
-    line = re.sub(r"[ù]", "u", line)
+    value = re.sub(r"[¢£¤¥§©ª®°²³µ¶¹º¼½¾×‰€™]", "-", value)
+    value = re.sub(r"[éè]", "e", value)
+    value = re.sub(r"[à]", "a", value)
+    value = re.sub(r"[ù]", "u", value)
     # changing unwanted character to "-"
-    line = re.sub(r"&\w+;", "-", line)
-    line = re.sub(
+    value = re.sub(r"&\w+;", "-", value)
+    value = re.sub(
         r"[\s!\"#\$%&'()*+,\/:;<=>?@\[\\\]^_`{\|}~¡¢£¤¥¦§¨©ª«¬®¯°±²³´µ¶·¸¹º»¼½¾¿×ˆ˜–—‘’‚“”„†‡•…‰‹›€™\t]",  # noqa: E501
         "-",
-        line,
+        value,
     )
-    # lowering the line if wanted
+    # lowering the value if wanted
     if lowercase:
-        line = line.lower()
+        value = value.lower()
     # removing excess "-"
-    line = re.sub(r"-+", "-", line)
-    line = line.strip("-")
-    return line
+    value = re.sub(r"-+", "-", value)
+    value = value.strip("-")
+    return value
--- a/robotoff/off.py
+++ b/robotoff/off.py
@@ -504,17 +504,17 @@ def select_rotate_image(
 
 def normalizing(line, lowercase=True):
     # removing accents
-    line = re.sub(r"[¢£¤¥§©ª®°²³µ¶¹º¼½¾×‰€™]", '-', line)
-    line = re.sub(r"[éè]", 'e', line)
-    line = re.sub(r"[à]", 'a', line)
-    line = re.sub(r"[ù]", 'u', line)
+    line = re.sub(r"[¢£¤¥§©ª®°²³µ¶¹º¼½¾×‰€™]", "-", line)
+    line = re.sub(r"[éè]", "e", line)
+    line = re.sub(r"[à]", "a", line)
+    line = re.sub(r"[ù]", "u", line)
     # changing unwanted character to "-"
     line = re.sub(r"&\w+;", "-", line)
     line = re.sub(
-                r"[\s!\"#\$%&'()*+,\/:;<=>?@\[\\\]^_`{\|}~¡¢£¤¥¦§¨©ª«¬®¯°±²³´µ¶·¸¹º»¼½¾¿×ˆ˜–—‘’‚“”„†‡•…‰‹›€™\t]",  # noqa: E501
-                "-",
-                line,
-            )
+        r"[\s!\"#\$%&'()*+,\/:;<=>?@\[\\\]^_`{\|}~¡¢£¤¥¦§¨©ª«¬®¯°±²³´µ¶·¸¹º»¼½¾¿×ˆ˜–—‘’‚“”„†‡•…‰‹›€™\t]",  # noqa: E501
+        "-",
+        line,
+    )
     # lowering the line if wanted
     if lowercase:
         line = line.lower()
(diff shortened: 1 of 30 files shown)


robotoff/prediction/ocr/category_from_AOC.py
```diff
@@ -0,0 +1,84 @@
...
+
+logger = get_logger(__name__)
+
+
+def process_category(lang, match) -> Optional[str]:
+    """Function to process the name of the category extracted from the AOC
+    recognition. We want it to match categories found in the categories
+    taxonomy database. If nothing is found, we don't return the category."""
+
+    unchecked_category = str(lang) + ":" + match.group().replace(" ", "-")
```
alexgarel, line 19: If you use a named capturing expression (see comment about avoiding positive/negative  look behind, you can here use `match.group("category")`

tests/unit/prediction/ocr/test_category_from_AOC.py
```diff
@@ -0,0 +1,19 @@
...
+
+from robotoff.prediction.ocr.category_from_AOC import AOC_REGEX, find_category_from_AOC
+
+
+@pytest.mark.parametrize(
+    "text,value_tags",
+    [
+        ("Appellation Clairette de Die Controlée", ["fr:clairette-de-die"]),
+        ("Appellation Clairette de Die Protégée", ["fr:clairette-de-die"]),
+        ("Chinon appellation d'origine protégée", ["fr:chinon"]),
```
alexgarel, line 13: Add checks on spaces:

```suggestion
        ("Chinon appellation d'origine protégée", ["fr:chinon"]),
        ("Chinon    appellation\t d'origine\n protégée", ["fr:chinon"]),
```

alexgarel (commented):
Cool you're heading the right direction :-)

robotoff/prediction/ocr/category_from_AOC.py
```diff
@@ -0,0 +1,84 @@
...
+
+    checked_category = get_taxonomy("category").__getitem__(unchecked_category)
+
+    if checked_category is not None:
+        return unchecked_category
+
+    return None
```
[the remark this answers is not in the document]
GabrielBeFr: I agree it does look a bit stupid. 
But actually checked_category is a Taxonomy and we must return a string so I can't return checked_category itself.
That's why I say : 
- if our unchecked_category detected by AOP REGEX does exist among already existing categories, we return it
- else we can't return unchecked_category because it doesn't match an existing category, so we return None

robotoff/prediction/ocr/category_from_AOC.py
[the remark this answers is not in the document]
GabrielBeFr: Indeed...
Should I add it to the off.py myself then ?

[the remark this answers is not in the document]
alexgarel: :+1:

robotoff/prediction/ocr/category_from_AOC.py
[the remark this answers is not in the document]
alexgarel: oooh that's me, the stupid !
In didn't saw **un**checked_category

Your code was correct.

That said you should return the checked_category tag because in find_category_from_AOC you have to put the tag for this category not the variable string you found.

robotoff/prediction/ocr/category_from_AOC.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def process_category(lang, match) -> Optional[str]:
+    """Function to process the name of the category extracted from the AOC
+    recognition. We want it to match categories found in the categories
+    taxonomy database. If nothing is found, we don't return the category."""
+
+    unchecked_category = str(lang) + ":" + match.group().replace(" ", "-")
+
+    checked_category = get_taxonomy("category").__getitem__(unchecked_category)
```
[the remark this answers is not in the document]
GabrielBeFr: It looks like a taxonomy has no get attribute/function and, as get_taxonomy returns a taxonomy, I can't use .get.
Indeed, at the end checked_category is a taxonomy. 
Besides that's the reason I return unchecked_category : the value_tag attribute of prediction requires a string, not a taxonomy.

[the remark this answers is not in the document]
GabrielBeFr: I actually have to do checked_category = get_taxonomy("category").nodes.get(unchecked_category).
But it is still a taxonomy so I can't put it as it is in the value_tag attribute of the prediction.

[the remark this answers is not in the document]
alexgarel: `checked_category.id` is the tag, I think.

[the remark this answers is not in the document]
GabrielBeFr: Thanks !!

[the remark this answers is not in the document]
alexgarel: @GabrielBeFr regarding the "get" method you could have added it to Taxonomy implementation :-)

robotoff/off.py
```diff
@@ -500,3 +500,25 @@ def select_rotate_image(
 
     r.raise_for_status()
     return r
+
+
+def normalizing(line, lowercase=True):
```
alexgarel, line 505: I would be more explicit, I would call it `normalize_tag` (it's about tag)

And think about adding documentation !

Also you should rename "line" parameter to "value" maybe ?

```suggestion
def normalize_tag(line, lowercase=True):
    """given a value normalize it to a tag (as in taxonomies)
    
    This means removing accents, lowercasing, replacing spaces with dashes, etc..
    """
```

robotoff/prediction/ocr/category_from_AOC.py
```diff
@@ -0,0 +1,109 @@
...
+from robotoff.utils import get_logger
+
+from .dataclass import OCRField, OCRRegex, OCRResult, get_text
+
+logger = get_logger(__name__)
+
+
+def category_taxonomisation(lang, match) -> Optional[str]:
+    """Function to match categories detected via AOP REGEX with categories
+    taxonomy database. If no match is possible, we return None."""
```
alexgarel, line 16: style: normally, in docstring, the last triple quotes is on a new line, this is a convention
```suggestion
def category_taxonomisation(lang, match) -> Optional[str]:
    """Function to match categories detected via AOP REGEX with categories
    taxonomy database. If no match is possible, we return None.
    """
```

robotoff/prediction/ocr/category_from_AOC.py
```diff
@@ -0,0 +1,109 @@
...
+
+    unchecked_category = lang + normalizing(match.group("category"))
+
+    checked_category = get_taxonomy("category").nodes.get(unchecked_category)
```
alexgarel, line 20: We may add a TODO comment, because this way I think we are not trying to match with synonyms.
(for that we may have to create a utility function in Taxonomy).

robotoff/prediction/ocr/category_from_AOC.py
```diff
@@ -0,0 +1,109 @@
...
+                            predictor="regex",
+                            data={"text": match.group(), "notify": ocr_regex.notify},
+                        )
+                    )
+    return predictions
+
+
+"""This function returns a prediction of the category of the product
+by detecting an AOC syntax which allows an easy category
+prediction with REGEX"""
```
alexgarel, line 109: I think this should be above :-)

robotoff/prediction/ocr/core.py
```diff
@@ -12,6 +12,7 @@
 from robotoff.utils.types import JSONType
 
 from .brand import find_brands
+from .category_from_AOC import find_category_from_AOC
```
alexgarel, line 15: I think we shall call the module category and the function find_categories, for we only have one function by prediction type. Maybe in the future we want to detect using not only AOC.

alexgarel (changes requested):
Almost there, very good. Some small changes still.

robotoff/off.py
```diff
@@ -502,23 +502,27 @@ def select_rotate_image(
...
     if lowercase:
-        line = line.lower()
+        value = value.lower()
     # removing excess "-"
-    line = re.sub(r"-+", "-", line)
-    line = line.strip("-")
-    return line
+    value = re.sub(r"-+", "-", value)
+    value = value.strip("-")
+    return value
```
alexgarel, line 528: thank you for this change :-)

robotoff/prediction/ocr/category_from_AOC.py
```diff
@@ -76,7 +80,14 @@ def category_taxonomisation(lang, match) -> Optional[str]:
...
 
-def find_category_from_AOC(content: Union[OCRResult, str]) -> List[Prediction]:
+def find_category(content: Union[OCRResult, str]) -> List[Prediction]:
+    """This function returns a prediction of the category of the product
+    by detecting an AOP syntax which allows an easy category
+    prediction with REGEX. 
+    For now we are extracting categories via REGEX 
+    only thanks to an AOP syntax but we may find in the future 
+    other ways to get sure prediction of categories.
+    """
```
alexgarel: I won't focus on AOP here. Your function could work for other categories as well.

That's also why I proposed you to rename the file to category.py (and you should do it ;-))

The only thing which is specific to AOP is the Regexp you defined and their processing function.
So we can keep it generic, one might as well add another regexp and processing function to add categories about honey type…

alexgarel (commented):
I hope I'm not too picky but still a file rename and doc change is needed, to avoid confuse future contributors :-)

robotoff/prediction/ocr/category_from_AOC.py
alexgarel: ping

robotoff/prediction/ocr/category_from_AOC.py
GabrielBeFr: Indeed ! I should have removed the mention of AOP in the docstring. Regarding the file name, that's weird because I have it under the name category.py in my local repository.... I'll commit it again then !

alexgarel (approved):
LGTM.

But you have a check failing, so you must resolve it before merging.

Issue #143, opened 2020-04-19
Add MSC or ASC as a parent when a MSC or ASC code is added


teolemon:
Add MSC or ASC as a parent when a MSC or ASC code is added

alexgarel:
@teolemon I'm a bit lost because the ticket description is a bit cryptic.

The scope is just to add the asc or msc label on the product right ?

(what is this notion of "parent" in the ticket ?)

PR #893, opened 2022-09-11
refactor: adding a test for MSC code

Jagrutiti:
### What
- Started with test cases to detect the code using the following sample data:

```
postgres=# select * from prediction where type='packager_code' and value like 'ASC%' order by id des
c limit 4;
    id    |    barcode    |     type      |                            data                            |         timestamp          | value_tag |    value    |      source_image       | automatic_processing |     server_domain     | predictor 
----------+---------------+---------------+------------------------------------------------------------+----------------------------+-----------+-------------+-------------------------+----------------------+-----------------------+-----------
 21205099 | 00355902      | packager_code | {"raw": "asc-c-00654", "type": "fishing", "notify": false} | 2022-09-06 10:58:38.528205 |           | ASC-C-00654 | /00355902/3.jpg         | t                    | api.openfoodfacts.org | flashtext
 21204554 | 00355902      | packager_code | {"raw": "asc-c-00654", "type": "fishing", "notify": false} | 2022-09-06 10:47:35.039248 |           | ASC-C-00654 | /00355902/2.jpg         | t                    | api.openfoodfacts.org | flashtext
 21143566 | 4316268664431 | packager_code | {"raw": "asc-c-00518", "type": "fishing", "notify": false} | 2022-09-03 16:02:28.009742 |           | ASC-C-00518 | /431/626/866/4431/2.jpg | t                    | api.openfoodfacts.org | flashtext
 21103197 | 0218978911697 | packager_code | {"raw": "asc-c-02452", "type": "fishing", "notify": false} | 2022-09-01 20:04:20.922998 |           | ASC-C-02452 | /021/897/891/1697/1.jpg | t                    | api.openfoodfacts.org | flashtext
```

### Fixes bug(s)
- #143


tests/unit/insights/test_importer.py
```diff
@@ -27,6 +27,17 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _set_up_and_tear_down(peewee_db):
+    # clean db
+    clean_db()
+    # Run the test case.
+    yield
+    clean_db()
+
```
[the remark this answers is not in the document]
alexgarel: So I don't think you need it.

tests/unit/insights/test_importer.py
```diff
@@ -629,6 +640,53 @@ def generate_insights(
         batch_insert_mock.assert_called_once()
         product_insight_delete_mock.assert_called_once()
 
+    def test_import_msc_insights(self, mocker):
+        class FakeImporter(InsightImporter):
```
alexgarel, line 644: I don't think you need a FakeImporter for your test.
You are not testing the mechanics of the Genral Importer, you are testing the PackagerCodeImporter.

tests/unit/insights/test_importer.py
```diff
@@ -629,6 +640,53 @@ def generate_insights(
...
+                        data={"raw": "asc-c-00518", "type": "fishing", "notify": False},
+                    )
+                ]
+
+        imported1 = FakeImporter.import_insights(
+            [Prediction(type="packager_code", value="ASC-C-00654")],
+            DEFAULT_SERVER_DOMAIN,
+            automatic=True,
+            product_store=FakeProductStore(),
+        )
```
alexgarel, line 677: you don't need FakeImporter but want to use PackagerCodeInsightImporter

Moreover if you call import_insights it's hard to have a unit test, as it is writing to the database,
but if you look at it, you will see that it gets insights from generate_insights and that's what is important to you (you want to see if the right insights are generated), that's the function most of the test are testing in this file :-)

Your current test is inside the TestInsightImporter which is testing the generic InsightImporter, that's why it creates FakeImporters (because InsightImporter is not complete per se). This is not what you are testing.

Look instead, in this file there is already a TestPackagerCodeInsightImporter and it has a test_generate_candidates method !

So you just need to make a second method like this :-)
(or parametrize the existing test)

alexgarel (commented):
@Jagrutiti some explanations.

alexgarel (commented):
Well done. I'm not approving yet, since the test was a preparatory work :-)

alexgarel:
Merging as it adds comments and a test.

Issue #466, opened 2021-10-11
Add support to flashtext USDA packager codes


teolemon:
### What

- https://github.com/openfoodfacts/openfoodfacts-server/issues/3704
- The list is available in the XLSX, in the second tab: "Est. 23466", but there are probably variants we'll find on OCRs.
<img src="https://user-images.githubusercontent.com/1689815/85283489-bcd9cc00-b48d-11ea-8057-1e074d2a8fc6.png" height="400px">
<img src="https://world.openfoodfacts.org/images/products/089/310/100/1041/front_en.5.400.jpg" height="400px">

### Part of
- #374 
- [US support](https://github.com/openfoodfacts/openfoodfacts-server/issues/5689)

GabrielBeFr:
I pushed my modifications to solve this problem about a week ago but, after executing the ocr prediction on a dump of ocr results, I realised the amount of false positive was way too high. I will change my REGEX to make the search more precise. I will also put the Automatic processing to False in order to avoid any problem if  we were to put this new feature in prod. 
@teolemon what do you think of limitating the search of USDA packager codes to US products ? Do you know if non-US product can also have USDA packager codes ?

PR #897, opened 2022-09-13
feat: Extract USDA packager codes with REGEX and flashtext

GabrielBeFr:
### What
- This PR is solving the issue https://github.com/openfoodfacts/robotoff/issues/466, allowing Robotoff to detect USDA packager codes.
- I added a data file containing all the USDA codes that exist according to the United States Department of Agriculture. It is read by the file packager_code.py via a flashtext to match possible USDA codes detected via REGEX with existing codes.
- I thus added a few functions in the file packager_code.py involved in the process_func associated to USDA codes in the REGEX dict.

### Part of 
- https://github.com/openfoodfacts/robotoff/issues/466


robotoff/prediction/ocr/packager_code.py
```diff
@@ -34,75 +34,136 @@ def process_fsc_match(match) -> str:
...
-    "fsc": OCRRegex(
-        re.compile(r"fsc.? ?(c\d{6})"),
-        field=OCRField.text_annotations,
-        lowercase=True,
-        processing_func=process_fsc_match,
-    ),
-    "eu_fr": OCRRegex(
-        re.compile(
-            r"fr (\d{2,3}|2[ab])[\-\s.](\d{3})[\-\s.](\d{3}) (ce|ec)(?![a-z0-9])"
+def process_USDA_match_to_flashtext(match) -> Optional[Any]:
```
alexgarel, line 37: It's a bit lazy to use Any here ;-)
Isn't it Optional[str] ?

robotoff/prediction/ocr/packager_code.py
```diff
@@ -34,75 +34,136 @@ def process_fsc_match(match) -> str:
...
+
+    unchecked_code = match.group().upper()
+    unchecked_code = re.sub(r"\s*\.*", "", unchecked_code)
+
+    processor = USDA_CODE_KEYWORD_PROCESSOR_STORE.get()
+    USDA_code = extract_USDA_code(processor, unchecked_code)
+    return USDA_code
+
+
+def generate_USDA_code_keyword_processor() -> KeywordProcessor:
```
alexgarel, line 49: Please add a docstring to every function :-)
Eg.
```suggestion
def generate_USDA_code_keyword_processor() -> KeywordProcessor:
    """Build the KeyWordProcessor for USDA codes
    
    This will be called only once thanks to CachedStore
    """
```

robotoff/prediction/ocr/packager_code.py
```diff
@@ -34,75 +34,136 @@ def process_fsc_match(match) -> str:
...
+
+    codes = text_file_iter(settings.OCR_USDA_CODE_FLASHTEXT_DATA_PATH)
+    return generate_keyword_processor(codes)
+
+
+def extract_USDA_code(processor: KeywordProcessor, text: str) -> Optional[Any]:
```
alexgarel, line 55: Given a string return the USDA code it contains (thanks to the processor) or None

tests/unit/prediction/ocr/test_packager_code.py
```diff
@@ -15,6 +15,9 @@
         ("FR-AB0-123", []),
         ("fr-098-123", []),
         ("Gluten code is FR-234-234 ", ["FR-234-234"]),
+        ("EST  \n 31778", ["EST. 31778"]),
+        ("EST  \n 9999", []),
+        ("M31779+  P31779+  \tV31779", ["M31779 + P31779 + V31779"]),
```
alexgarel, line 20: We should not have this, instead we should have:
```suggestion
        ("M31779+  P31779+  \tV31779", ["M31779", "P31779", "V31779"]),
```

For this test to pass you will need to change your code.

For more information, see PackagerCodeAnnotator and how emb_code is a list of codes.

robotoff/prediction/ocr/packager_code.py
```diff
@@ -34,75 +34,136 @@ def process_fsc_match(match) -> str:
...
-                        "raw": match.group(0),
-                        "type": regex_code,
-                        "notify": ocr_regex.notify,
-                    },
-                    type=PredictionType.packager_code,
-                    automatic_processing=True,
-                )
-            )
+    for regex_code, regex_list in PACKAGER_CODE.items():
+        for ocr_regex in regex_list:
```
alexgarel, line 142: cool to have changed the structure this way :-)

robotoff/prediction/ocr/packager_code.py
```diff
@@ -34,75 +34,136 @@ def process_fsc_match(match) -> str:
...
+        # To match the USDA like "EST. 522"
+        OCRRegex(
+            re.compile(r"EST\.*\s*\d{1,5}[A-Z]{0,3}\.*"),
+            field=OCRField.full_text_contiguous,
+            lowercase=False,
+            processing_func=process_USDA_match_to_flashtext,
+        ),
+        # To match the USDA like "V34626" or "M34614 + P34614 + V34614"
+        OCRRegex(
+            re.compile(r"[A-Z]\d{1,5}[A-Z]?(\s*\+\s*[A-Z]\d{1,5}[A-Z]?){0,3}"),
```
alexgarel, line 129: Do we really need to capture the " + " in one go, and not just each code ? This would simplify the task I think.

robotoff/prediction/ocr/packager_code.py
```diff
@@ -34,75 +34,136 @@ def process_fsc_match(match) -> str:
...
+
+    codes = text_file_iter(settings.OCR_USDA_CODE_FLASHTEXT_DATA_PATH)
+    return generate_keyword_processor(codes)
+
+
+def extract_USDA_code(processor: KeywordProcessor, text: str) -> Optional[Any]:
+    USDA_code = None
+
+    for (USDA_code_keyword, _) in processor.extract_keywords(text):
+        USDA_code = USDA_code_keyword
```
alexgarel, line 59: It's a bit strange to have a for loop without an explanation
Either you `break` on first value (explicit)
Or you add a comment on why it is safe.

alexgarel (changes requested):
Good work, still there is a edge case which is not handled correctly.

robotoff/prediction/ocr/packager_code.py
GabrielBeFr: Ok, I wasn't sure. I must admit that I am not really used to types. Thanks !

robotoff/prediction/ocr/packager_code.py
GabrielBeFr: Ok will do

robotoff/prediction/ocr/packager_code.py
GabrielBeFr: Actually, I did this for the same reason I did what you saw as a mistake two comments upwards. I thought that some products had codes which were explicitly "M34614 + P34614 + V34614". That's what written in the data @teolemon put in the [Add support for the USDA packager codes #3704](https://github.com/openfoodfacts/openfoodfacts-server/issues/3704), [Available as an XLSX](https://docs.google.com/spreadsheets/d/1W7Z7R_wc5Rj9AabCr5hk5-JHgi-cJCxO/edit#gid=381973512).

alexgarel (approved):
Thanks ! LGTM.

alexgarel:
@GabrielBeFr you have a flake8 issue to resolve before we merge.

alexgarel:
@teolemon and @raphael0202 I have one question: on this kind of matching on a packager code, which is quite simple, don't we want to limit to either product with english text, or even products in the US ? (or North America)

alexgarel:
:tada: @GabrielBeFr :-)

Issue #901, opened 2022-09-14
Allow filtering unanswered questions by product country


alexfauquette:
The idea would be to ask hunger game users which country interest them and display the top question for their country

alexgarel:
@Jagrutiti would you like to tackle this one while it's fresh ? :-)

Jagrutiti:
We can add a country parameter to the `UnansweredQuestionCollection` API.

The good thing is `get_insights` already has a country parameter to filter. We only have to pass it through the API.

Once you confirm the solution I can start working on it. @alexgarel

alexgarel:
Yes it's as you said @Jagrutiti !

raphael0202:
Fixed

PR #905, opened 2022-09-20
feat: Filter product by country to get answered questions

Jagrutiti:
### What
- Added `country `parameter to the `UnansweredQuestionCollection` API.

### Closes
- #901

### DIFF
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -1158,7 +1158,7 @@ class UnansweredQuestionCollection:
         pdb.set_trace()
 
         query_parameters = {
-            "keep_types": [question_type],
+            "keep_types": [question_type] if question_type else None,
             "group_by_value_tag": True,
             "value_tag": value_tag,
             "limit": count,
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -1150,13 +1150,17 @@ class UnansweredQuestionCollection:
         count: int = req.get_param_as_int("count", min_value=1, default=25)
         question_type: str = req.get_param("type")
         value_tag: str = req.get_param("value_tag")
+        country: Optional[str] = req.get_param("country")
         server_domain: Optional[str] = req.get_param("server_domain")
+
+        import pdb; pdb.set_trace()
 
         query_parameters = {
             "keep_types": [question_type],
             "group_by_value_tag": True,
             "value_tag": value_tag,
             "limit": count,
+            "country": country,
             "server_domain": server_domain,
         }
 
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -1153,7 +1153,9 @@ class UnansweredQuestionCollection:
         country: Optional[str] = req.get_param("country")
         server_domain: Optional[str] = req.get_param("server_domain")
 
-        import pdb; pdb.set_trace()
+        import pdb
+
+        pdb.set_trace()
 
         query_parameters = {
             "keep_types": [question_type],
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -1153,10 +1153,6 @@ class UnansweredQuestionCollection:
         country: Optional[str] = req.get_param("country")
         server_domain: Optional[str] = req.get_param("server_domain")
 
-        import pdb
-
-        pdb.set_trace()
-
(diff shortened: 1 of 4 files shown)


robotoff/app/api.py
```diff
@@ -1150,13 +1150,19 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
...
         value_tag: str = req.get_param("value_tag")
+        country: Optional[str] = req.get_param("country")
         server_domain: Optional[str] = req.get_param("server_domain")
 
+        import pdb
+
+        pdb.set_trace()
+
         query_parameters = {
             "keep_types": [question_type],
```
alexgarel, line 1161: Replace this by:
```suggestion
            "keep_types": [question_type] if question_type else None,
```

alexgarel: That said, I don't think using this API without question_type as any sense, so I would better make question_type mandatory.

robotoff/app/api.py
```diff
@@ -1150,13 +1150,19 @@ def on_get(self, req: falcon.Request, resp: falcon.Response):
         count: int = req.get_param_as_int("count", min_value=1, default=25)
         question_type: str = req.get_param("type")
         value_tag: str = req.get_param("value_tag")
+        country: Optional[str] = req.get_param("country")
         server_domain: Optional[str] = req.get_param("server_domain")
 
+        import pdb
+
+        pdb.set_trace()
+
```
raphael0202, line 1159: Can you remove this debugging code?

robotoff/app/core.py
```diff
@@ -100,6 +100,8 @@ def get_insights(
     if keep_types:
         where_clauses.append(ProductInsight.type.in_(keep_types))
 
+    # import pdb; pdb.set_trace()
+
```
raphael0202, line 104: Same comment as above about debugging

tests/integration/test_api.py
```diff
@@ -655,6 +655,26 @@ def test_get_unanswered_questions_api(client):
     assert data["status"] == "found"
 
 
+def test_get_unanswered_questions_api_with_country_filter(client):
+    ProductInsight.delete().execute()  # remove default sample
+
+    # test for filter with "country"
+
+    d = ProductInsightFactory(
```
raphael0202, line 663: You don't have to assign the product to `d` variable as you're not using it (local variable 'd' is assigned to but never used)

raphael0202 (changes requested):

tests/integration/test_core_integration.py
```diff
@@ -172,30 +172,33 @@ def test_get_unanswered_questions_list():
...
     insight_data5 = get_insights(keep_types=["category"])
     insight_data_items5 = [item.to_dict() for item in insight_data5]
     assert len(insight_data_items5) == 2
 
+    product6 = ProductInsightFactory(value_tag="en:raisins", countries="en:india")
+    insight_data6 = get_insights(country="en:india")
+    insight_data_items6 = [item.to_dict() for item in insight_data6]
+    assert insight_data_items6[0]["id"] == product6.id
+    assert insight_data_items6[0]["value_tag"] == "en:raisins"
+    assert insight_data_items6[0]["countries"] == "en:india"
```
alexgarel, line 200: you should also test you have only 1 product

```suggestion
    assert len(insight_data_items6) == 1
```

tests/integration/test_api.py
```diff
@@ -655,6 +655,26 @@ def test_get_unanswered_questions_api(client):
...
 
 
+def test_get_unanswered_questions_api_with_country_filter(client):
+    ProductInsight.delete().execute()  # remove default sample
+
+    # test for filter with "country"
+
+    ProductInsightFactory(
+        type="location", value_tag="en:dates", barcode="032", countries=["en:india"]
+    )
```
alexgarel, line 665: Adding only one product, you don't really test the filter is effective.

To test that you should add another product in another country
```suggestion
    ProductInsightFactory(
        type="location", value_tag="en:dates", barcode="033", countries=["en:france"]
    )
```

alexgarel (approved):
We need just a small improvement on tests.

I validate the PR but please, apply suggestions before merging, if you agree.

PR #906, opened 2022-09-21
refactor: robustness and localdev

alexgarel:
### What

- make some part of the code and settings more compliant to a use with a local deploy of product opener
- settings had a dangerous pattern in the object to get domains, fixed it.
- also made scheduler more resilient to not having network 

Note: a more serious refactor of networks between docker should be done, but later...

This was done to be able to test https://github.com/openfoodfacts/openfoodfacts-server/pull/7379

### DIFF
--- a/.env
+++ b/.env
@@ -18,7 +18,7 @@ ROBOTOFF_DOMAIN=openfoodfacts.localhost
 
 # for dev we change static to .net
 STATIC_OFF_DOMAIN=https://openfoodfacts.net
-# if you want to connect to a producopener dev instance on localhost, use:
+# if you want to connect to a Product Opener dev instance on localhost, use:
 # STATIC_OFF_DOMAIN=http://openfoodfacts.localhost
 # for dev scheme is http
 ROBOTOFF_SCHEME=http
--- a/.env
+++ b/.env
@@ -15,8 +15,11 @@ PO_LOCAL_NET=po_default
 TAG=latest
 ROBOTOFF_INSTANCE=localhost
 ROBOTOFF_DOMAIN=openfoodfacts.localhost
+
 # for dev we change static to .net
 STATIC_OFF_DOMAIN=https://openfoodfacts.net
+# if you want to connect to a producopener dev instance on localhost, use:
+# STATIC_OFF_DOMAIN=http://openfoodfacts.localhost
 # for dev scheme is http
 ROBOTOFF_SCHEME=http
 # for dev only on localhost
@@ -54,6 +57,6 @@ SENTRY_DSN=
 
 # Workers
 IPC_AUTHKEY=ipc
-IPC_HOST=0.0.0.0
+IPC_HOST=workers
 IPC_PORT=6650
 WORKER_COUNT=8
--- a/docker/dev.yml
+++ b/docker/dev.yml
@@ -44,6 +44,9 @@ services:
       po_local:
         aliases:
         # trick: make robotoff available outside and inside with same name
+        # with this alias, we can call it from inside docker with same name as outside
+        # and Product Opener will call it from a browser (outside)
+        # or from within perl code (inside)
         - robotoff.openfoodfacts.localhost
         - api
       webnet:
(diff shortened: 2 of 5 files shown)


robotoff/scheduler/__init__.py
```diff
@@ -202,11 +203,13 @@ def _refresh_elasticsearch():
...
 def _update_data():
     """Refreshes the PO product dump and updates the Elasticsearch index data."""
-
-    _download_product_dataset()
-    # Elasticsearch is dependent on the availability of the PO product dump, i.e.
-    # it it called after the download product dataset call.
-    _refresh_elasticsearch()
+    try:
+        _download_product_dataset()
+        # Elasticsearch is dependent on the availability of the PO product dump, i.e.
```
teolemon, line 208:
```suggestion
        # Elasticsearch is dependent on the availability of the Product Opener product dump, i.e.
```

alexgarel:
```suggestion
        # Elasticsearch is dependent on the availability of the product dump from Product Opener 
        # (main Open Food Facts backend) 
```

robotoff/scheduler/__init__.py
```diff
@@ -202,11 +203,14 @@ def _refresh_elasticsearch():
...
-    # Elasticsearch is dependent on the availability of the PO product dump, i.e.
-    # it it called after the download product dataset call.
-    _refresh_elasticsearch()
+    try:
+        _download_product_dataset()
+        # Elasticsearch is dependent on the availability of the product dump from Product Opener
+        # (main Open Food Facts backend)
+        # it it called after the download product dataset call.
+        _refresh_elasticsearch()
+    except requests.exceptions.ConnectionError:
```
raphael0202, line 212: I would prefer using `requests.exceptions.RequestException` to catch any type of request errors instead of just a `ConnectionError`

.env
```diff
@@ -15,8 +15,11 @@ PO_LOCAL_NET=po_default
 TAG=latest
 ROBOTOFF_INSTANCE=localhost
 ROBOTOFF_DOMAIN=openfoodfacts.localhost
+
 # for dev we change static to .net
 STATIC_OFF_DOMAIN=https://openfoodfacts.net
+# if you want to connect to a producopener dev instance on localhost, use:
```
raphael0202, line 21:
```suggestion
# if you want to connect to a Product Opener dev instance on localhost, use:
```

docker/dev.yml
```diff
@@ -31,16 +31,22 @@ x-robotoff-dev: &robotoff-dev
...
+      webnet:
 
 services:
   api:
     <<: *robotoff-dev
-    <<: *networks-productopener-local
+    networks:
+      po_local:
+        aliases:
+        # trick: make robotoff available outside and inside with same name
```
raphael0202, line 46: Can you explain in more details how it works?

robotoff/off.py
```diff
@@ -125,7 +125,8 @@ def get_api_product_url(server: Union[ServerType, str]) -> str:
 def get_base_url(server: Union[ServerType, str]) -> str:
     if isinstance(server, str):
         server = server.replace("api", "world")
-        return "https://{}".format(server)
+        scheme = settings.BaseURLProvider().scheme
```
raphael0202, line 128: It would be good to add a comment to explain why this is useful here (to be able to work with a local Product Opener in dev mode)

raphael0202 (changes requested):
See attached comments

docker/dev.yml
alexgarel: I added `# with this alias, we can call it from inside with same name as outside`

An alias is an additional name you give to the container in this network. Here I give it the name "robotoff.openfoodfacts.localhost". I need this because on Product Opener I have only one config which is ROBOTOFF_URL and it is used by the javascript (so in the browser, which is not in docker space) and by the perl code (which is inside docker). So I need a name which work for both. And the easiest way is this :-)

alexgarel:
I've done requested changes @raphael0202

raphael0202 (approved):

Issue #833, opened 2022-07-22
TestLabelInsightImporter should not depend on external data


alexgarel:
### What

TestLabelInsightImporter depends on taxonomies served by openfoodfacts.
This has several drawbacks:
1. it might prevent a coder to test his code because of low internet bandwidth or the .net server being down (happens to @Jagrutiti )
2. it makes the test fails without warning and without any change to the code, on some taxonomy changes (happens this week) 

### Expected behavior

Test should only depend on data provided with the code.

We should have a simple json for each taxonomy we use in tests with several categories, that we could use it to monkeypatch `robotoff.taxonomy.get_taxonomy`.

### parts of

#649

raphael0202:
We already have a JSON for each taxonomy locally (which is the fallback version in case of network error)

alexgarel:
@raphael0202 you have it if you're not on a fresh install, here running test eg. in CI means you are on a fresh install. (Or I am wrong ?)

raphael0202:
Even on a fresh install the taxonomies are available here: https://github.com/openfoodfacts/robotoff/tree/master/data/taxonomies
But I just noticed a bug for brand taxonomy, it fallbacks by default (after request timeout) to the label one

PR #909, opened 2022-09-22
refactor: fix test_importer.py unit tests

raphael0202:
### What
Fixes issue #833.
Taxonomy files have been updated as well (some test cases didn't pass on LabelInsightImporter because one label was not in the taxonomy).


alexgarel (approved):
Cool :-)

alexgarel:
@raphael0202 do not forget to use semantic titles :-)

As fix / feat will appear in release, I prefer to use a refactor for tests (although we may customize prefixes and add test: prefix)

alexgarel:
And I merged with another commit name, as you did indeed fix something !

PR #912, opened 2022-09-23
feat: expose postgres DB locally

raphael0202:
Goal: being able to connect to production database from local computer to:
- get stats
- debugging purpose


alexgarel (approved):
I propose a different default, as I prefer not to expose ports if not needed (security).

PR #914, opened 2022-09-26
fix: store category neural model in ProductInsight.predictor field

raphael0202:
data->model is a deprecated field, all predictor version/type must be stored in the predictor field


alexgarel (approved):
Good move :-)

PR #915, opened 2022-09-27
fix: remove trailing slash in all URLs

raphael0202:
For API consistency. It's not a breaking change, calling these routes with trailing slash still works.


alexgarel (approved):
:+1:

PR #925, opened 2022-09-30
fix: Stopping tests when one already failed

GabrielBeFr:
### What
- Adding the -x parameter to the run pytest command for integration test in the Makefile. 

### Why
- When running integration tests, we'd like to stop the tests if one is failing in order to prevent a too large amount of errors from being displayed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,7 +146,7 @@ integration-tests:
 	@echo "🥫 Running integration tests …"
 	# run tests in worker to have more memory
 	# also, change project name to run in isolation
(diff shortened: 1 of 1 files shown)


alexgarel:
I'm not really in favor of it personally, I prefer to see all failing. Some time a second test gives you a better understanding of what's going on.

raphael0202:
Locally that's what we usually want, but not on CI pipeline, when it's interesting to know all tests that failed.

raphael0202:
There is actually an interactive command you can use I didn't know of: `make pytest args='-x'`

Jagrutiti:
Sometimes, one change results in the failure of two or more test cases.

I would like to see all the failed tests, to fix them at once rather than one by one.

PR #936, opened 2022-10-03
fix: don't display questions about insights that are automatically applicable

raphael0202:
There is no need to display these questions as the insight is applied automatically max 5 minutes later.


alexgarel (approved):
LGTM

(maybe adding a test ?)

raphael0202:
It would be good to test this huge function indeed

PR #937, opened 2022-10-03
feat: Improve predict category endpoint

raphael0202:
- feat: add a `threshold` parameter to /predict/category endpoint
- feat: Improve /predict/category endpoint (allow to provide model input instead of only fetching the product from
DB)
- doc: document /predict/category endpoint on OpenAPI

### DIFF
--- a/robotoff/app/schema.py
+++ b/robotoff/app/schema.py
@@ -1,6 +1,6 @@
-from typing import Any, Dict
+from robotoff.utils.types import JSONType
 
-IMAGE_PREDICTION_IMPORTER_SCHEMA: Dict[str, Any] = {
+IMAGE_PREDICTION_IMPORTER_SCHEMA: JSONType = {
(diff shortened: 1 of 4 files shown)


robotoff/app/schema.py
```diff
@@ -40,3 +40,48 @@
...
+                },
+                "deepest_only": {
+                    "type": "boolean",
+                },
+                "threshold": {"type": "number"},
+            },
+            "required": ["product"],
+        },
+    ],
+}
```
alexgarel, line 87: It's a bit of a pity we cannot reuse the api.yml.

I'm asking myself if it would be the case to load api.yml at module load time, and reuse sub parts ?

alexgarel (approved):
Good move :-)

robotoff/app/schema.py
raphael0202: Indeed it's a bit of code duplication... Most routes however don't accept JSON as content

PR #938, opened 2022-10-04
fix: only save annotation_result in ProductInsight when needed

raphael0202:
No need to save it if the provided data was invalid or if it's just a vote

### DIFF
--- a/robotoff/insights/annotate.py
+++ b/robotoff/insights/annotate.py
@@ -36,13 +36,13 @@ class AnnotationResult:
 class AnnotationStatus(Enum):
     saved = 1
     updated = 2
-    vote_saved = 9
     error_missing_product = 3
     error_updated_product = 4
     error_already_annotated = 5
     error_unknown_insight = 6
     error_missing_data = 7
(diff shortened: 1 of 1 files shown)


robotoff/insights/annotate.py
```diff
@@ -125,8 +125,15 @@ def _annotate(
...
             annotation_result = SAVED_ANNOTATION_RESULT
 
-        insight.annotated_result = annotation_result.status_code
-        insight.save()
+        if annotation_result.status_code in (
+            AnnotationStatus.saved.value,
+            AnnotationStatus.updated.value,
+            AnnotationStatus.error_invalid_image.value,
+            AnnotationStatus.error_missing_product.value,
+            AnnotationStatus.error_updated_product.value,
```
alexgarel, line 133: Style: you could have made a "db_status" attribute to AnnotationStatus, to have here a simpler:   `annotation_result.status_code in AnnotationStatus.db_status`

alexgarel (approved):
I propose a style change but otherwise it's perfectly ok.

PR #944, opened 2022-10-07
fix: adding documentation about the insights/annotate.py file

GabrielBeFr:
### What
- Adding documentation in insights/annotate.py about what it is used for and how to use it.


robotoff/insights/annotate.py
```diff
@@ -23,6 +23,18 @@
 from robotoff.products import get_image_id, get_product
 from robotoff.utils import get_logger
 
+"""
+This file allows to annotate product.
```
raphael0202, line 27:
```suggestion
This file allows to annotate products.
```

robotoff/insights/annotate.py
```diff
@@ -23,6 +23,18 @@
...
+
+To check whether the annotation already exists or not, save it and send it to the off database, use the following commands :
```
raphael0202, line 29:
```suggestion
To check whether the annotation already exists or not (and save it and send it to the off database), use the following commands:
```

robotoff/insights/annotate.py
```diff
@@ -23,6 +23,18 @@
 from robotoff.products import get_image_id, get_product
 from robotoff.utils import get_logger
 
+"""
+This file allows to annotate products.
+
+To check whether the annotation already exists or not (and save it and send it to the off database), use the following commands:
```
teolemon, line 29:
```suggestion
To check whether the annotation already exists or not (and save it and send it to the Open Food Facts database), use the following commands:
```

robotoff/insights/annotate.py
```diff
@@ -23,6 +23,18 @@
...
+    from robotoff.insights.annotate import InsightAnnotatorFactor
+    annotator = InsightAnnotatorFactory.get(insight_type)
+    annotator.annotate(insight: ProductInsight, annotation: int, update: bool = True, data: Optional[Dict] = None, auth: Optional[OFFAuthentication] = None, automatic: bool = False)
+
+If you don't want to add it to the off database but only save it for example in case of a vote, you can use SAVED_ANNOTATION_VOTE_RESULT to warn the annotation has been taken into account.
```
teolemon, line 34:
```suggestion
If you don't want to add it to the Open Food Facts (off) database but only save it for example in case of a vote, you can use SAVED_ANNOTATION_VOTE_RESULT to warn the annotation has been taken into account. 
```

robotoff/insights/annotate.py
```diff
@@ -23,6 +23,18 @@
...
 
+"""
+This file allows to annotate products.
+
+To check whether the annotation already exists or not (and save it and send it to the Open Food Facts database), use the following commands:
+    from robotoff.insights.annotate import InsightAnnotatorFactor
+    annotator = InsightAnnotatorFactory.get(insight_type)
+    annotator.annotate(insight: ProductInsight, annotation: int, update: bool = True, data: Optional[Dict] = None, auth: Optional[OFFAuthentication] = None, automatic: bool = False)
+
+If you don't want to add it to the Open Food Facts (off) database but only save it for example in case of a vote, you can use SAVED_ANNOTATION_VOTE_RESULT to warn the annotation has been taken into account.
```
raphael0202, line 34:
```suggestion
If you don't want to update the Open Food Facts database but only save the insight annotation (if the update is performed on the client side for example), you can call `annotate()` with `update=False`.
```

raphael0202 (approved):

PR #950, opened 2022-10-11
fix: Update generate-docs.yml

sumit-158:


raphael0202:
Well spotted, thanks!

PR #956, opened 2022-10-13
feat: improve logo annotation route

raphael0202:
- feat: support providing taxonomized value as input during logo annotation
- feat: improve request validation during logo annotation

### DIFF
--- a/robotoff/app/api.py
+++ b/robotoff/app/api.py
@@ -764,6 +764,7 @@ class ImageLogoDetailResource:
 
 
 class ImageLogoAnnotateResource:
+    @jsonschema.validate(schema.ANNOTATE_LOGO_SCHEMA)
     def on_post(self, req: falcon.Request, resp: falcon.Response):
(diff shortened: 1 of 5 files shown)


robotoff/app/api.py
```diff
@@ -776,19 +777,29 @@ def on_post(self, req: falcon.Request, resp: falcon.Response):
...
+                )
 
             logo.annotation_type = type_
             logo.username = username
             logo.completed_at = completed_at
-            logo.save()
             annotated_logos.append(logo)
 
+        for logo in annotated_logos:
+            logo.save()
```
GabrielBeFr, line 801: Why doing another loop ?

GabrielBeFr (commented):
Looks good to me, just have one question

GabrielBeFr (approved):

robotoff/app/api.py
raphael0202: So that we check that every logo is correct before saving them in DB.
Previously we could save some of the and throw an HTTP 400 error.

PR #963, opened 2022-10-17
fix: fix insight import mecanism

raphael0202:
The main addition of this PR is a fix of the insight import mecanism.

Currently, when a prediction based on a more recent image is available, we don't transform this prediction into an insight, we keep the old insight.
This is problematic as the insight based on the newer image may make it automatically processable (if the image is selected or if it's not too old compared to most recent images). As knowing which insight to keep/delete and which candidate to import is tricky (many subtle bugs were discovered since the introduction of this feature), I chose here a more radical approach: we regenerate all insights for a product from predictions and only keep annotated insights in DB.

I also added a better sorting of candidates:
- we now set a higher priority on candidates that are automatically processable (this field may be set after the ProductInsight creation, that's why we cannot sort according to this field in `sort_predictions`)
- we now set a higher priority on candidates that have a non null predictor field

### DIFF
--- a/robotoff/cli/main.py
+++ b/robotoff/cli/main.py
@@ -17,10 +17,11 @@ def run(service: str) -> None:
 
 
 @app.command()
-def generate_import_insights(
+def regenerate_ocr_insights(
     barcode: str = typer.Argument(..., help="Barcode of the product")
 ) -> None:
-    """Generate OCR predictions/insights for a specific product and import them."""
+    """Regenerate OCR predictions/insights for a specific product and import
+    them."""
     from robotoff import settings
     from robotoff.insights.extraction import (
         DEFAULT_OCR_PREDICTION_TYPES,
--- a/robotoff/cli/main.py
+++ b/robotoff/cli/main.py
@@ -21,15 +21,15 @@ def generate_import_insights(
     barcode: str = typer.Argument(..., help="Barcode of the product")
 ) -> None:
     """Generate OCR predictions/insights for a specific product and import them."""
+    from robotoff import settings
     from robotoff.insights.extraction import (
         DEFAULT_OCR_PREDICTION_TYPES,
         extract_ocr_predictions,
     )
     from robotoff.insights.importer import import_insights as import_insights_
     from robotoff.off import generate_json_ocr_url
-    from robotoff.utils import get_logger
-    from robotoff import settings
     from robotoff.products import get_product
+    from robotoff.utils import get_logger
 
     logger = get_logger()
 
@@ -296,11 +296,9 @@ def refresh_insights(
     refreshed, otherwise insights of all products are refreshed.
     """
     from robotoff import settings
+    from robotoff.insights.importer import refresh_all_insights
+    from robotoff.insights.importer import refresh_insights as refresh_insights_
     from robotoff.utils import get_logger
-    from robotoff.insights.importer import (
-        refresh_insights as refresh_insights_,
-        refresh_all_insights,
-    )
 
     logger = get_logger()
     server_domain = server_domain or settings.OFF_SERVER_DOMAIN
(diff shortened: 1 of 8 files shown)


raphael0202:
The OCR category prediction commit fixes this bug that has been introduced in latest release:
https://sentry.io/organizations/openfoodfacts/issues/3536599898/?project=1415205&query=is%3Aunresolved
https://sentry.io/organizations/openfoodfacts/issues/3671738567/?project=1415205&query=is%3Aunresolved

robotoff/insights/importer.py
```diff
@@ -311,8 +343,9 @@ def generate_insights(
             for candidate in candidates:
                 if candidate.automatic_processing is None:
                     logger.warning(
-                        "Insight with automatic_processing=None: %s", candidate.__data__
+                        "Insight with automatic_processing=None", candidate.__data__
```
alexgarel, line 346: This way candidate.__data__ won't be used !
Either you want to log it and need %s
Either you just want it in sentry and you can add it with `extra={"canditate": candidate.__data__}`

robotoff/insights/importer.py
```diff
@@ -375,13 +408,16 @@ def get_insight_update(
...
+            for reference in (
+                reference_insight
+                for reference_insight in reference_insights
+                if reference_insight.annotation is not None
+            ):
                 if cls.is_conflicting_insight(candidate, reference):
                     # Candidate conflicts with existing insight, keeping
                     # existing insight and discarding candidate
-                    to_keep_ids.add(reference.id)
                     match = True
```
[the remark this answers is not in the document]
alexgarel: I would even do both check (against reference_insights and to_create) in one go, without any performance penalty:
```python
match = any(
    cls.is_conflicting_insight(candidate, reference_insight)
    for reference_insight in itertools.chain(reference_insights, to_create)
     if reference_insight.annotation is not None
)
```

robotoff/products.py
```diff
@@ -22,6 +22,12 @@
 
 
 def get_image_id(image_path: str) -> Optional[str]:
+    """Return the image ID from an image path.
+
+    :param image_path: the image path (ex: /322/247/762/7888/2.jpg)
+    :return: the image ID ("2" in the previous example) or None if the image
+    is not "raw" (not digit-numbered)
+    """
```
alexgarel, line 30: Cool to document on the go :-)

robotoff/cli/main.py
```diff
@@ -17,25 +17,38 @@ def run(service: str) -> None:
 
 
 @app.command()
-def predict_insight(ocr_url: str) -> None:
-    import json
-
+def generate_import_insights(
+    barcode: str = typer.Argument(..., help="Barcode of the product")
+) -> None:
+    """Generate OCR predictions/insights for a specific product and import them."""
```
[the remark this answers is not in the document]
raphael0202: Indeed, I've renamed the command `regenerate_ocr_insights` as it's only based on OCR currently.

robotoff/insights/importer.py
raphael0202: Well it did render with '%s' in sentry, that's why. But it's really minor, I revert this.

robotoff/insights/importer.py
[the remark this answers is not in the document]
raphael0202: Good idea for the use of `any`! The second option introduces a bug, as candidates in `to_create` all have `annotation=None` (as their candidate). So I will keep the first option :)

robotoff/insights/importer.py
alexgarel: FYI: It's the normal pattern for sentry, and that's for sentry that we use` %s", param`, because it helps sentry group entries.

PR #966, opened 2022-10-18
fix: various fixes

raphael0202:
- fix: don't return auto processable insights in /questions/{barcode}
- fix: fix /logos/search route
- feat: improve taxonomized value matching
- fix: fix `get_tag` function
- fix: update tags of some labels detected with flashtext
- fix: rename image_url field into ref_image_url
- test: add unit test for CategoryQuestionFormatter
- chore: add test and style semantic PR types
- doc: fix /predict/category description


robotoff/insights/question.py
```diff
@@ -240,10 +242,7 @@ class LabelQuestionFormatter(QuestionFormatter):
 
     def format_question(self, insight: ProductInsight, lang: str) -> Question:
         value_tag: str = insight.value_tag
-        image_url = None
-
-        if value_tag in LABEL_IMAGES:
-            image_url = LABEL_IMAGES[value_tag]
+        ref_image_url = LABEL_IMAGES.get(value_tag)
```
GabrielBeFr, line 245: In the loaded dictionnary LABEL_IMAGES, is there all the labels in the taxonomy ? Because if it's not the case, how do we handle the case where value_tag is not among the dictionnary ?

robotoff/models.py
```diff
@@ -72,7 +72,7 @@ class ProductInsight(BaseModel):
     """Table to store the insights generated by Robotoff."""
 
     # ID is the unique ID for this insight.
-    id = peewee.UUIDField(primary_key=True)
+    id = peewee.UUIDField(primary_key=True, default=uuid.uuid4)
```
GabrielBeFr, line 75: What is this doing ?

tests/unit/test_taxonomy.py
```diff
@@ -107,31 +102,20 @@ def test_find_deepest_nodes(
...
-    [
-        (TaxonomyType.brand.name, "carrefour-bio", "Carrefour Bio"),
-        (TaxonomyType.brand.name, "unknown-brand", None),
-        (TaxonomyType.label.name, "eu-organic", "en:eu-organic"),
+        (TaxonomyType.label.name, "fr:bio-europeen", "en:eu-organic"),
         (
             TaxonomyType.label.name,
             "ab-agriculture-biologique",
-            "fr:ab-agriculture-biologique",
+            None,
```
GabrielBeFr, line 109: Isn't it a downgrade not to allow ab-agriculture-biologique to match taxonomy ?

robotoff/insights/question.py
raphael0202: no there are only a few labels that have a logo. If the label ID is not in `LABEL_IMAGES`, None is passed as `ref_image_url`.

robotoff/models.py
raphael0202: Setting a default value to UUID id when creating a ProductInsight

tests/unit/test_taxonomy.py
raphael0202: we now only accept language-prefixed values for logo, so no it's expected!

Issue #622, opened 2022-02-14
Switch from Tensorflow Serving to Triton


mithridatea:
We currently use Tensorflow Serving to serve all models:
- object detection models  (nutriscore, universal-logo,...)
- category classifier model

However, Tensorflow Serving only supports the SavedModel format, which is specific to Tensorflow. It means we cannot easily integrate models trained with other frameworks (pytorch, scikit-learn,...).

ONNX is a standard format to save model architecture and weights supported by all major frameworks, including Tensorflow, Pytorch and scikit-learn.

## Proposal

Test Triton, an open-source inference server developed by Nvidia. Supported formats include TF SavedModel and ONNX. Runtime on both GPU and CPU.
If tests are successful, switch from Tensorflow Serving to Triton server.

ram690514:
I am interested to work on this issue , if no one is currently working on it

raphael0202:
The switch to Triton is done in preprod, closing this issue.

Issue #565, opened 2022-01-27
Should container-deploy-ml be triggered by master and deploy-*


alexgarel:
robotoff-ml is a production instance which is also used by preprod.

Actual definition of container-deploy-ml makes it being deployed on each preprod release.
Something is wrong there, and there is the risk of messing up robotoff prod !

So:
* either we deploy robotoff-ml only on v*.*.* tags (like prod)
* or we should have two distinct instances (one for preprod, one for prod)

alexgarel:
@mithridatea @CharlesNepote @ocervell your comments are welcome

alexgarel:
We could also avoid reloading models when there is no need of it, by putting a ML_VERSION file and running workflow only if this file changes.

mithridatea:
We rarely ship new ML models, so the simplest way to me is to trigger the deployment of TF Serving manually, when a new tag is added for instance (something like `tf-serving-deploy-*`).
About preprod/prod,  it would be best to have two distinct environments if we can afford it, so that we don't mess up production when testing a model on preprod (even if this case shouldn't occur, as we should always deploy a new TF Serving model version when updating a model).

alexgarel:
The idea of a new tag for deploying is simple and effective.

For prod / preprod split, I think we can afford it, for we gain on old robotoff instances.

raphael0202:
Models are now synced only when adding a version tag `v*.*.*` or when adding a tag `model-deploy*` (see #977)
Adding Triton to docker production server is still required, and can be done when we have enough RAM on this instance.

raphael0202:
TF Serving and Triton are now deployed in prod. robotoff-org (prod) uses these prod containers.

PR #977, opened 2022-10-28
feat: switch all object detection models to Triton (#622)

### LINKED ISSUE #942: Process crash when calling /predict/images with universal-logo-detector model
### What

API thread is most of the time killed (502 Bad Gateway) when `/predict/images` is called with `universal-logo-detector` or `nutrition-table` models.
https://robotoff.openfoodfacts.org/api/v1/images/predict?image_url=https://fr.openfoodfacts.org/images/products/801/973/005/1888/front_it.3.full.jpg&models=universal-logo-detector&output_image=1

The culprit is the size of the JSON response returned by TF Serving: 150 MB of text, 950 MB in RAM once deserialized.
In the response, we can find the features of the last layer of the network (which we don't need) that is really heavy.

This also has the downside of slowing down image processing and increasing memory usage in Robotoff workers.

Solution: export again the model as SavedModel without outputting this field, or find a way to query Tensorflow Serving to return predictions without this field.

raphael0202:
- switch to Triton for all object detection model, use gRPC protocol for faster transfer
- only request needed output fields (resolves #942)
- get rid of CategoryIndex and protobuf-generated files

I managed to migrate the category detector model to Triton (using the SavedModel asset) but didn't succeed in sending inference request, due to dimension issue: `ingredient` input shape is defined as (-1, ), `product_name` as (1, ). Triton requires all inputs to have the same number of elements for dim 0 (=batch_size), while it's not necessary in Tensorflow Serving as inputs are fed differently.
We could fix this dimension issue in the category predictor model code, but as we're going to change it anyway to a new model we can keep Tensorflow Serving to serve category prediction model in the meantime.


.github/workflows/models-deploy.yml
```diff
@@ -1,10 +1,10 @@
 name: Sync Tensorflow models to Proxmox VM
-on: 
+on:
   push:
     branches:
       - master
       - deploy-*
-    tags: 
+    tags:
       - v*.*.*
```
alexgarel, line 8: we have to change this to deploy on specific tags. see https://github.com/openfoodfacts/robotoff/issues/565

robotoff/prediction/object_detection/core.py
```diff
@@ -156,18 +197,18 @@ def get_available_models(cls) -> List[str]:
 
     @classmethod
     def load_all(cls):
```
alexgarel, line 199: I see this method is called at module importation (I find it unperfect already !).
Why do we do that, if it is meant to be served by triton ?

docker/ml.yml
```diff
@@ -14,5 +14,25 @@ services:
...
+    # - GRPC/HTTP support
+    # - CPU only (we don't have GPU in production)
+    # - Tensorflow 2 SavedModel and ONNX support
+    # This allows us to reduce significantly the image size
+    # See https://gist.github.com/raphael0202/091e521f2c79a8db8c6e9aceafb6e0b9 for build script
+    image: ghcr.io/openfoodfacts/triton:latest
+    ports:
+      - 8000:8000
+      - 8001:8001
+      - 8002:8002
```
alexgarel, line 29: I really would prefer if all port exposure where configurable through .env. 

See https://github.com/openfoodfacts/openfoodfacts-infrastructure/blob/develop/docs/docker.md#enable-configuration-through-environment

```suggestion
      - ${TRITON_EXPOSE_HTTP:-8000}:8000
      - ${TRITON_EXPOSE_GRPC:-8001}:8001
      - ${TRITON_EXPOSE_METRICS:-8001}:8002
```
In this case, it also has the advantage of documenting what ports are doing :blush:

alexgarel (approved):
I have some small comments, so I already approve, but it would be great to do changes.

robotoff/prediction/object_detection/core.py
raphael0202: We need the ID->label mapping (which is now in model directory in `labels.txt`), but it's indeed not necessary to load it at import time, we can load it lazily.

docker/ml.yml
raphael0202: It's a good practice indeed, I'm adding this

Issue #908, opened 2022-09-21
Make the skip mechanism only for user


alexgarel:
### Problem
When a user skip a question, it is no more available to anyone,
while most of the time user skip it only because he does not know enough on the category or has not enough bandwith for now.

### Proposed solution

Use AnnotationVote for that instead.
It seems to me it will work out of the box !
Instead of storing as an annotation the "skip", store it as an AnnotationVote.

### Part of
- #374

sumit-158:
@alexgarel Is it okk If I work on this? If yes, what is the target file?

alexgarel:
@raphael0202 is the one setting priorities ;-)

PR #986, opened 2022-11-08
fix: fix skip mechanism

raphael0202:
So that this mechanism does not have to be implemented client-side.
The voting mechanism has also been updated: before, all classes
(0, 1, -1) could be considered a valid value. Now, -1 is exclusively
used for users to ignore insights, so it's not considered when
computing vote counts (only 0 and 1 are considered).
Another bug was fixed: when vote counts for 0 and 1 are both >= 2,
we save the insight as invalid (-1) instead of saving it as negative
(0).

Solves #908

### DIFF
--- a/robotoff/app/core.py
+++ b/robotoff/app/core.py
@@ -287,10 +287,31 @@ def save_annotation(
     auth: Optional[OFFAuthentication] = None,
     trusted_annotator: bool = False,
 ) -> AnnotationResult:
-    """Saves annotation either by using a single response as ground truth or by using several responses.
+    """Saves annotation either by using a single response as ground truth or
+    by using several responses.
 
-    trusted_annotator: defines whether the given annotation comes from an authoritative source (e.g.
-    a trusted user), ot whether the annotation should be subject to the voting system.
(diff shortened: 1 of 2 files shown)


robotoff/app/core.py
```diff
@@ -332,24 +359,25 @@ def save_annotation(
...
-            existing_votes[0].num_votes >= 2
-            and len(existing_votes) > 1
-            and existing_votes[1].num_votes >= 2
-        ):
-            # This code credits the last person to contribute a vote with a potentially not their annotation.
-            annotation = 0
-            verified = True
+        if existing_votes:
+            # If the top annotation has more than 2 votes, consider applying it to the insight.
+            if existing_votes[0].num_votes > 2:
```
alexgarel, line 364: I think it would be more clear to use "3 votes or more" 
As I read the code, the "strictly more" was not obvious, and I was asking myself, how could the second case arise.
```suggestion
            # If the top annotation has 3 or more votes, consider applying it to the insight.
            if existing_votes[0].num_votes >= 3:
```

alexgarel (approved):
LGTM, cool !

PR #988, opened 2022-11-09
fix: fix insight import (again!)

### LINKED ISSUE #975: Voting mechanism is broken after insight import update
PR #963 introduced a new insight import mechanism that prevented insights based on old images from being kept in DB while new data is available. All non-annotated insights are recreated everytime the product is refreshed. This breaks the current voting mechanism, as insights that AnnotationVote refer may not exist anymore.

We should transfer votes to newly created insights of the same type/value.

raphael0202:
Make voting system work again, and avoid the deletion/creation of insights when there are only minor updates.
Also delete the legacy `automatic` parameter in InsightImporter.

Fixes #975

### DIFF
--- a/robotoff/insights/importer.py
+++ b/robotoff/insights/importer.py
@@ -290,7 +290,7 @@ class InsightImporter(metaclass=abc.ABCMeta):
                 raise ValueError(f"unexpected prediction type: '{prediction.type}'")
 
         inserts = 0
-        for to_create, to_delete in cls.generate_insights(
+        for to_create, to_update, to_delete in cls.generate_insights(
             predictions, server_domain, automatic, product_store
         ):
             if to_delete:
@@ -306,6 +306,9 @@ class InsightImporter(metaclass=abc.ABCMeta):
                     50,
                 )
 
+            for insight in to_update:
+                insight.save()
+
         return inserts
 
     @classmethod
@@ -315,9 +318,11 @@ class InsightImporter(metaclass=abc.ABCMeta):
         server_domain: str,
         automatic: bool,
         product_store: DBProductStore,
-    ) -> Iterator[Tuple[List[ProductInsight], List[ProductInsight]]]:
+    ) -> Iterator[
+        Tuple[List[ProductInsight], List[ProductInsight], List[ProductInsight]]
+    ]:
         """Given a list of predictions, yield tuples of ProductInsight to
-        create and delete.
+        create, update and delete.
 
         It calls the `generate_candidates` method, specific to each insight type
         (and implemented in sub-classes).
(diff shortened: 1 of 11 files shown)


robotoff/insights/importer.py
```diff
@@ -421,21 +444,53 @@ def get_insight_update(
...
+                    mapping_ref_insight = None
+                    # In order for the voting system to work, we map insights to create
+                    # to existing insights with the same value/value_tag/source_image.
+                    # This way, we don't loose vote information.
+                    for reference_insight in reference_insights:
+                        if (
+                            reference_insight.annotation is None
+                            and cls.is_conflicting_insight(candidate, reference_insight)
+                            # only map to existing insight is the source image is the same,
+                            # otherwise create a new insight
```
alexgarel, line 468:
```suggestion
                            # only map to existing insight if the source image is the same,
                            # otherwise create a new insight
```

robotoff/insights/importer.py
```diff
@@ -421,21 +444,53 @@ def get_insight_update(
...
+                            and candidate.source_image == reference_insight.source_image
+                        ):
+                            mapping_ref_insight = reference_insight
+                            to_keep_ids.add(reference_insight.id)
+                            break
```
alexgarel, line 473: Really picky: this is a good example where you could have use a `for…else` and put `mapping_ref_insight = None` in the else block, because you always break if you find.

tests/unit/insights/test_importer.py
```diff
@@ -278,23 +279,23 @@ def test_get_insight_update_annotated_references(self):
...
                 type=InsightType.label,
                 value_tag="tag1",
-                id=uuid.UUID("a6aa784b-4d39-4baa-a16c-b2f1c9dac9f9"),
                 annotation=-1,
             ),
             ProductInsight(
                 barcode=DEFAULT_BARCODE,
                 type=InsightType.label,
                 value_tag="tag2",
-                id=uuid.UUID("f3fca6c5-15be-4bd7-bd72-90c7abd2ed4c"),
```
alexgarel, line 288: Oooh yes you got the courage to remove this :+1:

alexgarel (approved):
LGTM

alexgarel:
Also cool you took the time to remove the legacy automatic attribute !

tests/unit/insights/test_importer.py
raphael0202: yes it's a good opportunity to do some cleanup :wink:

tests/integration/insights/test_category_import.py
```diff
@@ -104,7 +103,7 @@ def test_import_one_same_value_tag(self, predictions):
         """Test when there is a single import, but the value_tag stays the
         same."""
         imported = self._run_import(predictions)
-        assert imported == 1
+        assert imported == 0
         # no insight created
         assert ProductInsight.select().count() == 1
         insight = ProductInsight().select().limit(1)[0]
```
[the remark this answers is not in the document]
raphael0202: Indeed, good idea!

PR #1007, opened 2022-12-02
fix: add various fixes and improvements

raphael0202:
- fix: update barcode range check during insight import
- fix: fix default value for MONGO_URI in .env
- fix: fix mongoDB healthcheck
- fix: add lower value for MongoDB serverSelectionTimeoutMS 30s is too high
- fix: add mongodb service in webnet network
- fix: don't build robotoff image by default
- docs: fix error in dev-install.md
- feat: remove update_recycle task


Makefile
```diff
@@ -55,9 +55,9 @@ up:
 	@echo "🥫 Building and starting containers …"
 	docker network create po_default || true  
 ifdef service
-	${DOCKER_COMPOSE} up -d --build ${service} 2>&1
+	${DOCKER_COMPOSE} up -d ${service} 2>&1
```
alexgarel, line 58: :+1:

docker/mongodb.yml
```diff
@@ -8,6 +8,8 @@ services:
       - "127.0.0.1:${MONGO_EXPOSE_PORT:-27017}:27017"
     volumes:
       - mongodb_data:/data/db
+    networks:
+      - webnet
```
alexgarel, line 12: Don't do that, it is dangerous as it will potentially conflict on prod servers.

Why should you need that ?

raphael0202: Otherwise how would other services could access mongodb service if they don't share a common network?
And I'm not worried about the production settings, as it's a dev-only file.

alexgarel: When you launch a service, docker-compose creates a default network to which every service is connected. It's named ${COMPOSE_PROJECT_NAME}_default.

But ok, I see that we manually specified `webnet` for now for all services in docker-compose (this is bad, but we shall fix it later).

So ok for that, for now… we will have to fix this network thing in the future.

See https://github.com/openfoodfacts/openfoodfacts-infrastructure/issues/146 (FYI)

.env
```diff
@@ -58,7 +58,7 @@ INFLUXDB_AUTH_TOKEN=
 # INFLUXDB_HOST=host.docker.internal
 
 # MongoDB (dev setting, using product opener network)
-MONGO_URI=mongodb://mongodb.po_default:27017
```
alexgarel, line 61: can we keep that as a comment for those who have a local product opener instance ?

PR #1011, opened 2022-12-08
feat: add rq worker

raphael0202:
- use rq and redis queues to send tasks from API to workers
- use redis locks to avoid concurrent product insight updates or concurrent product_update tasks
- simplify job workflows
- save nutriscore and nutrition-table object detection model results in `image_prediction` table


robotoff/workers/tasks/import_image.py
```diff
@@ -47,37 +68,95 @@ def run_import_image_job(
...
+        image_url=image_url,
+        server_domain=server_domain,
+    )
+    enqueue_job(
+        run_nutrition_table_object_detection,
+        high_queue,
+        barcode=barcode,
+        image_url=image_url,
+        server_domain=server_domain,
+    )
```
alexgarel, line 94: (not sure) Don't we add more chance of having conflicts ? 
Shan't we instead have some dependency between them ?

robotoff/workers/tasks/import_image.py
```diff
@@ -47,37 +68,95 @@ def run_import_image_job(
...
+        and prediction.type == PredictionType.label
+        for prediction in predictions
+    ):
+        enqueue_job(
+            run_nutriscore_object_detection,
+            high_queue,
+            barcode=barcode,
+            image_url=image_url,
+            server_domain=server_domain,
+        )
```
alexgarel, line 124: I would set a boolean variable to True and launch the job at the end.

Doing that you augment the risk of conflict (even with a lock, it's better to avoid it).

robotoff/workers/tasks/import_image.py
```diff
@@ -47,37 +68,95 @@ def run_import_image_job(
...
+        if product is None:
+            continue
+        save_image(barcode, source_image, product, server_domain)
 
 
 def save_image(
     barcode: str, source_image: str, product: Product, server_domain: str
 ) -> Optional[ImageModel]:
     """Save imported image details in DB."""
+    if existing_image_model := ImageModel.get_or_none(source_image=source_image):
```
alexgarel, line 154: Morse operator \o/

robotoff/workers/tasks/import_image.py
```diff
@@ -47,37 +68,95 @@ def run_import_image_job(
...
+    """Save a batch of images in DB.
+
+    :param batch: a batch of (barcode, source_image) tuples
+    :param server_domain: the server domain to use
+    """
+    for barcode, source_image in batch:
+        product = get_product_store()[barcode]
+        if product is None:
+            continue
+        save_image(barcode, source_image, product, server_domain)
```
alexgarel, line 147: Do you have really big batches ?
If yes, I would really try to fetch existing image in one request, then process the addition in one go (bulk insert).
If you do this one by one, appart that it takes a longer time, you augment the risk of having the whole transaction fail because one image added in between…

Another option to keep code simple, is to use a transaction by save_image (instead of setting it on whole operation).

robotoff/workers/tasks/import_image.py
```diff
@@ -126,57 +205,155 @@ def save_image(
...
+            item for item in image_prediction.data["objects"] if item["score"] >= 0.5
+        ]
+
+        if len(results) > 1:
+            logger.info("more than one nutriscore detected, discarding detections")
+            return
+
+        result = results[0]
+        score = result["score"]
+        label_tag = NUTRISCORE_LABELS[result["label"]]
```
alexgarel, line 271: Why put it in the `with db`, it shan't it be outside ?

robotoff/workers/tasks/import_image.py
```diff
@@ -126,57 +205,155 @@ def save_image(
...
-                image_prediction=image_prediction,
-                index=i,
-                score=item["score"],
-                bounding_box=item["bounding_box"],
+        if image_prediction is None:
+            # Can occur in normal conditions if an image prediction
+            # already exists for this image and model
+            return
+
+        logo_ids = []
```
alexgarel, line 320: This again, should not be in `with db` (if I don't miss something)

robotoff/app/api.py
```diff
@@ -458,22 +468,19 @@ def on_post(self, req: falcon.Request, resp: falcon.Response):
...
-                "image_url": image_url,
-                "ocr_url": ocr_url,
-                "server_domain": server_domain,
-            },
+        high_queue.enqueue(
+            run_import_image_job,
+            barcode=barcode,
+            image_url=image_url,
+            ocr_url=ocr_url,
+            server_domain=server_domain,
```
alexgarel, line 482: shouldn't you use enqueue_job to be consistent ?

robotoff/workers/tasks/product_updated.py
```diff
@@ -15,21 +16,35 @@
...
+    logger.info(f"Running `update_insights` for product {barcode} ({server_domain})")
 
-    updated_product_predict_insights(barcode, product_dict, server_domain)
-    logger.info("Refreshing insights...")
-    imported = refresh_insights(barcode, server_domain)
-    logger.info(f"{imported} insights created after refresh")
+    try:
+        with Lock(
+            name=f"robotoff:product_update_job:{barcode}", expire=300, timeout=10
+        ):
```
alexgarel, line 33: I think it's worth commenting 
```suggestion
        # We handle concurrency thanks to the lock
        # as the task will fetch product from MongoDB at the time it runs,
        # it's not worth reprocessing with another task arriving concurrently.
        # The expire is there only in case the lock is not released (process killed)
        with Lock(
            name=f"robotoff:product_update_job:{barcode}", expire=300, timeout=10
        ):
```

alexgarel (commented):
LGTM, here are my comments.

Only the change in deploy is potentially harmfull, we must discuss it or revert.

doc/how-to-guides/deployment/maintenance.md
```diff
@@ -47,7 +47,7 @@ robotoff_api_1         /bin/sh -c /docker-entrypo ...   Up      0.0.0.0:5500->55
                                                                 /tcp                                
 robotoff_postgres_1    docker-entrypoint.sh postg ...   Up      127.0.0.1:5432->5432/tcp            
 robotoff_scheduler_1   /bin/sh -c /docker-entrypo ...   Up                                          
-robotoff_workers_1     /bin/sh -c /docker-entrypo ...   Up                              
+robotoff_worker_low_1     /bin/sh -c /docker-entrypo ...   Up
```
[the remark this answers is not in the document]
raphael0202: I'm only listing a few ones for conciseness, but I'm adding robotoff_worker_high_1 :+1:

robotoff/cli/main.py
```diff
@@ -188,31 +213,180 @@ def refresh_insights(
...
+                job_kwargs={"result_ttl": 0},
+                batch=batch,
+                server_domain=settings.OFF_SERVER_DOMAIN,
+            )
+
+
+class ObjectDetectionModel(enum.Enum):
+    nutriscore = "nutriscore"
+    universal_logo_detector = "universal-logo-detector"
+    nutrition_table = "nutrition-table"
```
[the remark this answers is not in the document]
raphael0202: Discussed during a meeting, we should add all types/enum in a separate file to avoid duplicates (the goal was to allow CLI to load fast)

robotoff/scheduler/__init__.py
```diff
@@ -44,38 +44,39 @@
...
+                )
+                annotation_result = annotator.annotate(insight, 1, update=True)
+                processed += 1
+
+                if annotation_result == UPDATED_ANNOTATION_RESULT and insight.data.get(
+                    "notify", False
+                ):
+                    slack.NotifierFactory.get_notifier().notify_automatic_processing(
+                        insight
+                    )
```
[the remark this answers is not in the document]
raphael0202: So that even if there is one error during the annotation of one product (caused by Product Opener timeout for example), we still keep the annotation status for those that have been annotated.

robotoff/workers/tasks/import_image.py
raphael0202: I added the following comment in the code:
```
    # The two following tasks take longer than the previous one, so it
    # shouldn't be an issue to launch tasks concurrently (and we still have
    # the insight import lock to avoid concurrent insight import in DB for the
    # same product)
```

raphael0202: And it's interesting to launch them concurrently so that we get logo/nutriscore predictions faster.

robotoff/workers/tasks/import_image.py
raphael0202: I will switch to a transaction in `save_image`. The idea of this function is to have a simple way to import all images that were not imported due to some failure (see `import_images_in_db` CLI command)

robotoff/workers/tasks/import_image.py
raphael0202: `run_object_detection_model` creates a ImagePrediction instance, so we need the DB connection.
But it's indeed better to split it in 2, insight import failure shouldn't mean the ImagePrediction is not imported.

robotoff/app/api.py
raphael0202: yes totally!

robotoff/scheduler/__init__.py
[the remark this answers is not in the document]
alexgarel: If you put it in db.atomic() you will have this more surely

PR #1041, opened 2023-01-03
docs: ann documentation in robotoff technical reference

GabrielBeFr:
### What
- Adding docs for logos embedding and ANN


doc/research/logo-detection/ann-benchmark.md
```diff
@@ -34,7 +34,7 @@ HNSW indexes are adjustable trhough three parameters:
 
 ## Benchmark
 
-You can find the code of the various benchmarks [here](https://github.com/openfoodfacts/openfoodfacts-ai/tree/607ec6a/logo-ann/benchmark-ann).
```
raphael0202: it's better to keep a reference to a specific commit, so that if we update the code we know which version was used to generate the benchmark

raphael0202 (changes requested):

doc/research/logo-detection/ann-benchmark.md
```diff
@@ -0,0 +1,88 @@
...
+
+- (faiss)[https://github.com/facebookresearch/faiss]
+- (redis)[https://redis.io/docs/stack/search/reference/vectors/]
+- (elasticsearch)[https://www.elastic.co/fr/blog/introducing-approximate-nearest-neighbor-search-in-elasticsearch-8-0]
+
+For each framework, multiple parameters were used to find the best index in our use case.
+
+## Our Use Case
+
+The data used to test the performance of the indexes was made of 4.371.343 vectors (each one issued from the CLIP vision model with a logo as input) of dimension 768 and type float32.
```
raphael0202, line 15: We should add a paragraph to say we used the wrong embedding for the benchmark (size 768 instead of 512), but that it shouldn't change the result as there is only an additional projection. And that we use embedding of size 512 in production.

raphael0202 (changes requested):

doc/references/logos-ANN.md
```diff
@@ -0,0 +1,40 @@
+# Logo-ANN
+
+About 1600 products are added every day to the database.
+Each prodcuts having multpile logos on its packaging, thousands of new logos are added each day to Open Food Facts.
```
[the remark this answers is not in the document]
raphael0202: "are added each day to Open Food Facts" -> it's a weird formulation, logos are not really added.

raphael0202 (approved):

PR #1050, opened 2023-01-24
feat: implement packaging parsing from OCR texts

raphael0202:
Implement a packaging parser using lark grammar
Only support French for the moment.


robotoff/prediction/ocr/grammar.py
```diff
@@ -0,0 +1,155 @@
...
+    seen_set: dict[str, str] = {}
+    previous_node_id = None
+
+    node_id_names = extract_taxonomy_names(
+        lang, taxonomy, ignore_ids, add_synonyms, add_xx, lowercase, strip_accent
+    )
+    for node_id, name in node_id_names:
+        if node_id != previous_node_id:
+            if name in seen_set:
+                error_message = f"item {name} (node {node_id}) already seen for node {seen_set[name]}!"
```
alexgarel, line 128: I don't really understand this, could you add a comment to explain ? (I don't understand why we only check this if `node_id != previous_node_id`)

robotoff/prediction/ocr/grammar.py
```diff
@@ -0,0 +1,155 @@
...
+                terminal_priority if terminal_priority is not None else ""
+            )
+            item = f"{terminal_name}.{terminal_priority_str}: "
+            first_node = False
+            len_first_item = len(item) - 2
+        else:
+            item = f"{' ' * len_first_item}| "
+
+        name = name.replace("/", "\\/")
+        item += f'/(?<!\\w){name}(?!\\w)/  // "{node_id}"'
```
alexgarel, line 151: Don't you want to use word boundaries instead of look ahead and look behind which can be expensive.

```suggestion
        item += f'/\b{name}\b/  // "{node_id}"'
```

robotoff/prediction/ocr/grammar.py
```diff
@@ -0,0 +1,155 @@
...
+        texts.append(item)
```
alexgarel, line 152: This need some comments !

robotoff/prediction/ocr/grammar.py
```diff
@@ -0,0 +1,155 @@
...
+    terminal_name: str,
+    taxonomy_type: TaxonomyType,
+    terminal_priority: Optional[int] = None,
+    lowercase: bool = True,
+    strip_accent: bool = True,
+    add_synonyms: bool = True,
+    add_xx: bool = True,
+    ignore_ids: Optional[set[str]] = None,
+    raises: bool = True,
+) -> tuple[str, dict[str, list[str]]]:
```
alexgarel, line 115: A docstring would be really welcome.

robotoff/prediction/ocr/packaging.py
```diff
@@ -0,0 +1,217 @@
...
+    "en:net",  # filet
+    "en:pouch-flask",  # pouch, english word
+    "en:spoon",  # most of the spoon are false positive
+    "en:tablespoon",
+    "en:mold",  # moule
+    None,
+}
+
+
+def generate_packaging_lark_file(lang: str):
```
alexgarel, line 38:
```suggestion
def generate_packaging_lark_file(lang: str):
    """Generate the list of terminal symbols corresponding to taxonomies
    """
```

robotoff/prediction/ocr/packaging.py
```diff
@@ -0,0 +1,217 @@
...
+@cachetools.cached(cachetools.LRUCache(maxsize=5))
+def load_grammar(lang: str, start: str = "value", **kwargs) -> Lark:
+    return Lark.open(
+        str(settings.GRAMMARS_DIR / f"packaging_{lang}.lark"),
+        start=start,
+        **kwargs,
+    )
+
+
+class PackagingFRTransformer(Transformer):
```
alexgarel, line 66: Please add docstrings to explain a little bit.

robotoff/prediction/ocr/packaging.py
```diff
@@ -0,0 +1,217 @@
...
+        self.taxonomy_maps = taxonomy_maps
+
+    def value(self, items: list):
+        return items
+
+    def packaging(self, items: list):
+        data = {}
+        for item in items:
+            data.update(item)
+        return data
```
alexgarel, line 82: why not: 
```suggestion
        return dict(items)
```

robotoff/prediction/ocr/packaging.py
alexgarel:
```suggestion
class PackagingFRTransformer(Transformer):
    """For each processed rule, lark will call corresponding transformer method 
    with items composing it as parameters
    (which maybe themselves the result of previous rules)
    """
```

robotoff/prediction/ocr/packaging.py
```diff
@@ -0,0 +1,217 @@
...
+      `material` and `recycling`. Each field has a `value` and a `value_tag`
+      field. `value_tag` is None if the detected string could not be mapped
+      to the taxonomy value or if the string maps to more than one taxonomy
+      value (this shouldn't happen).
+    """
+    # Only fr is supported currently
+    lang = "fr"
+    grammar = load_grammar(lang)
+    processed_text = normalize_string(text, lowercase=True, strip_accent=True)
+    processed_text = strip_consecutive_spaces(processed_text)
```
alexgarel, line 174:
```suggestion
    processed_text = strip_consecutive_spaces(processed_text).strip()
```

robotoff/prediction/ocr/packaging.py
```diff
@@ -0,0 +1,217 @@
...
+def find_packaging(content: Union[OCRResult, str]) -> list[Prediction]:
+    text = get_text(content)
+    lang = "fr"
+    if match := match_packaging(text):
+        predictions = []
+        for item in match:
+            if "shape" not in item:
+                # No shape defined
+                # Safeguard, should not occur with current lark grammar
+                continue
```
alexgarel, line 193: maybe a log error ?

robotoff/prediction/ocr/packaging.py
```diff
@@ -0,0 +1,217 @@
...
+        ignore_ids={"en:unknown"},
+    )
+    generate_terminal_symbols_file(
+        output_path=settings.GRAMMARS_DIR / f"terminal_packaging_material_{lang}.lark",
+        lang=lang,
+        terminal_name="PACKAGING_MATERIALS",
+        taxonomy_type=TaxonomyType.packaging_material,
+        terminal_priority=1,
+        ignore_ids={"en:unknown"},
+    )
```
alexgarel, line 54: why don't we have recycling also ?

alexgarel (commented):
Really cool, I approve already, but I propose some cosmetic changes. Do not skip the one asking more comments :-)

alexgarel (approved):

data/grammars/terminal_packaging_material_fr.lark
```diff
@@ -0,0 +1,335 @@
+// This file has been generated automatically, DO NOT EDIT!
```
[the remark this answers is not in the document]
raphael0202: How is it use exactly?
Good to know anyway, it could be an issue in a near future

robotoff/insights/importer.py
```diff
@@ -490,6 +492,42 @@ def get_insight_update(
...
+            candidates,
+            key=lambda candidate: (
+                candidate.data.get("priority", 1),
+                -int(get_image_id(candidate.source_image) or 0)
+                if candidate.source_image
+                else 0,
+                # automatically processable insights come first
+                -int(candidate.automatic_processing),
+                # hack to set a higher priority to prediction with a predictor value
+                candidate.predictor or "z",
```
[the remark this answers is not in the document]
raphael0202: A sleeping emoji would be appropriate maybe? :laughing:

robotoff/insights/importer.py
```diff
@@ -900,6 +938,171 @@ def generate_candidates(
...
+        return {PredictionType.packaging}
+
+    @classmethod
+    def is_conflicting_insight(
+        cls, candidate: ProductInsight, reference: ProductInsight
+    ) -> bool:
+        # Only keep one insight per element (=shape)
+        return candidate.data["element"].get("shape", {}).get(
+            "value_tag"
+        ) == reference.data["element"].get("shape", {}).get("value_tag")
```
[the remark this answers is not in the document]
raphael0202: Not really... that's the limit of it.

robotoff/insights/importer.py
```diff
@@ -900,6 +938,171 @@ def generate_candidates(
...
+        predicted. The purpose of this method is to discard predicted
+        candidate elements that don't bring any information to the product.
+
+        Examples:
+        - candidate {shape: bottle} is uninformative if product already has an
+          element {shape: bottle, material: plastic}
+        - candidate {shape: bottle-cap, material: PET} is informative and
+          should be kept even if product already has an element
+          {shape: bottle-cap, material: plastic} (as plastic is a parent of
+          PET in material taxonomy)
```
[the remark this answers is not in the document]
raphael0202: if product has `{shape: cap, material: plastic}` we don't discard it

robotoff/prediction/ocr/packaging.py
raphael0202: Because recycling instructions were generated manually, to split between "VERB" and "NON_VERB".
It also required a few manual fixes. I've added a comment in the file to explain this.

robotoff/prediction/ocr/packaging.py
raphael0202: why this?

robotoff/prediction/ocr/packaging.py
raphael0202: It doesn't work, as each item needs to be a (k, v) tuple

robotoff/prediction/ocr/grammar.py
```diff
@@ -0,0 +1,155 @@
...
+            names += node.synonyms.get(lang, [])
+            if add_xx:
+                names += node.synonyms.get("xx", [])
+
+        names = [normalize_string(name, lowercase, strip_accent) for name in names]
+        # Remove empty items
+        names = [name for name in names if name.strip()]
+        output += [
+            (node.id, name) for name in sorted(set(names), key=lambda x: names.index(x))
+        ]
```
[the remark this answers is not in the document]
raphael0202: It's indeed costly, but for the task here we don't really care: the generation of terminal symbol is very rare, and it was almost instant on my laptop.

robotoff/prediction/ocr/grammar.py
raphael0202: Me neither :sweat_smile: 
This condition shouldn't exist, we should check if there is a match against any name seen so far

robotoff/prediction/ocr/grammar.py
raphael0202: I didn't know about `\b`, that's a nice trick, thanks!

robotoff/prediction/ocr/grammar.py
```diff
@@ -0,0 +1,155 @@
...
+            names.append(node.names["xx"])
+
+        if add_synonyms:
+            names += node.synonyms.get(lang, [])
+            if add_xx:
+                names += node.synonyms.get("xx", [])
+
+        names = [normalize_string(name, lowercase, strip_accent) for name in names]
+        # Remove empty items
+        names = [name for name in names if name.strip()]
```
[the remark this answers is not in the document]
raphael0202: Sure, we can do this as a safeguard!

robotoff/insights/importer.py
```diff
@@ -900,6 +938,171 @@ def generate_candidates(
...
+        product: Product,
+        predictions: list[Prediction],
+    ) -> Iterator[ProductInsight]:
+        # 1 prediction = 1 packaging element
+        for prediction in (
+            prediction
+            for prediction in predictions
+            if cls.keep_prediction(prediction, product)
+        ):
+            yield ProductInsight(**prediction.to_dict())
```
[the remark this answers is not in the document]
raphael0202: > Here if we have two prediction quite similar, we would emit two insight.

I prefer to emit a single insight at a time, and once the insight has been rejected/accepted, we refresh the insights for this (product, insight type). I know we do this when the insight has been accepted and a request was sent to Product Opener (as we get a product_updated webhook event), but I'm not sure we do it when the insight has been rejected.

> shouldn't we add existing insight to product packagings in keep_prediction

If I understand correctly what you suggest, it's already the case, as we provide all predictions for a product when calling `import_insights`, including the predictions that were used to generate the current insights

robotoff/insights/importer.py
```diff
@@ -900,6 +938,171 @@ def generate_candidates(
...
+          PET in material taxonomy)
+        """
+        for prop in PackagingElementProperty:
+            candidate_element.setdefault(prop.value, None)
+            ref_element.setdefault(prop.value, None)
+
+        if candidate_element["shape"] != ref_element["shape"]:
+            # The candidate element doesn't have the same shape as reference
+            # so it cannot be a superset
+            return False
```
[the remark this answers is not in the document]
raphael0202: There are a few cases where it is interested as well to take shape hierarchy into consideration, like:
candidate: {"shape": "en:box", "material": "en:ldpe-low-density-polyethylene"}
reference: {"shape": "en:pizza-box", "material": "en:plastic"}
Here the candidate refers to the same shape as the reference, even if it's called with a more generic name. So we keep the prediction.
I've added support for this + unit tests.

raphael0202:
Thank you for the review, I've added almost all the proposed changes, only a single non-mandatory one is pending. I'm merging the branch, but I will update the code afterwards if necessary after your comment.

PR #1051, opened 2023-01-24
fix: remove legacy code

raphael0202:
Remove unused code that generated "latent" insights.


GabrielBeFr (approved):
The function get_image_orientation() was indeed unused in robotoff and the logs modifications are fine.

PR #1057, opened 2023-02-02
fix: manually cropping and resizing of logos before CLIP processor

GabrielBeFr:
### What
- Adding manual resizing with PIL library.

### Why
- [CLIP processor]() was used as [the only processing step for logos](https://github.com/openfoodfacts/robotoff/blob/master/robotoff/triton.py#L23). However, to fit the size (224,224) expected by CLIP, it resizes only the smallest dimension of the image and then crops the logo. We don't want to loose data by cropping.


raphael0202 (approved):
LGTM!

PR #1060, opened 2023-02-16
ci: mongodb migration to off3

alexgarel:
We moved mongodb to a temporary server to be able to upgrade our servers.


raphael0202 (approved):
When should we deploy this?
edit: I just noticed the switch has been made manually on production server. Good for me!

PR #1064, opened 2023-03-08
chore: fix unit test action failure

raphael0202:
Fix long-standing failure in unit test & check GitHub action.
This issue arose a few months ago, probably after a change on Github side. The issue comes from permission issues on shared (mounted) folders between the host and the docker container: the process doesn't have write access on these shared folders, which is necessary to:

- compile i18n .po files (it was done before launching tests)
- write coverage reports in .cov folder

I didn't manage to solve the permission issue, but found a workaround:

- don't compile the po files before launching tests, as we already do that in Dockerfile
- run unit and integration tests inside the container as root user

This solution is not perfect (as we should avoid running processes as root), but it only affects GitHub actions.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,77 +30,28 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      # Taken from https://github.com/snok/install-poetry
-
       #----------------------------------------------
-      #       check-out repo and set-up python
+      #       check-out repo
       #----------------------------------------------
       - name: Check out repository
         uses: actions/checkout@v3
-      - name: Set up python
-        uses: actions/setup-python@v3
         with:
-          python-version: ${{ matrix.python-version }}
+          lfs: true
 
       #----------------------------------------------
-      #  -----  install & configure poetry  -----
-      #----------------------------------------------
-      - name: Install Poetry
-        uses: snok/install-poetry@v1.3
-        with:
-          virtualenvs-create: true
-          virtualenvs-in-project: true
-
-      #----------------------------------------------
-      #       load cached venv if cache exists
-      #----------------------------------------------
-      - name: Load cached venv
-        id: cached-poetry-dependencies
-        uses: actions/cache@v3
-        with:
-          path: .venv
-          key: venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
-
-      #----------------------------------------------
-      # install dependencies if cache does not exist
-      #----------------------------------------------
-      - name: Install dependencies
-        run: poetry install
-        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-
-      #----------------------------------------------
-      # Run linters
-      #----------------------------------------------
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -86,14 +86,11 @@ jobs:
...
       #----------------------------------------------
       - name: Install gettext
         run: sudo apt-get install gettext
 
-      - name: Compile translations
-        run: (cd i18n && bash compile.sh)
-
       #----------------------------------------------
       # Launch tests
       #----------------------------------------------
```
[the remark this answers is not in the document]
teolemon: ok, désolé @raphael0202 je teste des trucs :-)

alexgarel:
@raphael0202 normally the problem you describe arise because you mount a folder that does not exists. When it happens, docker creates a folder owned by root on the fly.

The normal way to fix it is to add an empty folder (with a .empty or .keepit file inside, so that it's in git) with right permissions.
It might be better to fix this also because

> run unit and integration tests inside the container as root user

This might hide bugs if you run it with the normal user.

raphael0202:
> The normal way to fix it is to add an empty folder (with a .empty or .keepit file inside, so that it's in git) with right permissions.

This issue does not come from missing directories, as i18n folder exists in the repository, and the translations could not get compiled (permission denied). This used to work correctly, I suspect an update from GitHub Action to have broken the pipeline.

Same thing for .cov folder, we create it in Dockerfile, to avoid permission issues.

alexgarel:
@raphael0202 I want to look into this before we merge.

alexgarel:
The bug is indeed strange to me.

> Same thing for .cov folder, we create it in Dockerfile, to avoid permission issues.
Yes but we mount it in dev.yml, so I though it was a problem with `OFF_UID` not being set, but it is indeed…

.github/workflows/pull_request.yml
```diff
@@ -30,77 +30,28 @@ jobs:
...
-        run: (cd i18n && bash compile.sh)
-
-      #----------------------------------------------
-      # Launch tests
+      # Launch checks and tests
       #----------------------------------------------
       - name: Configure docker
         run: |
+          # ensure a new line for .env file might not have it !
+          echo "" >> .env
```
alexgarel, line 47: @raphael0202 this was the error ! We loosed the new line at end of `.env`
So the `"OFF_UID=$UID"` was not written on a new line (but attached to the last) !
Thus the premissions problem (because runner user does not necessarily have user id 1000).

Makefile
```diff
@@ -163,11 +164,11 @@ docs:
 	@echo "🥫 Generationg doc…"
 	${DOCKER_COMPOSE} run --rm --no-deps api ./build_mkdocs.sh
 
-checks: toml-check flake8 black-check mypy isort-check docs
+checks: toml-check flake8 black-check mypy isort-check
```
alexgarel, line 167: why did you remove docs from checks ? Because they are built in another workflow ?

raphael0202:
Thank you for looking into this and fixing the issue @alexgarel! It's way cleaner this way :)
The "docs" removal was a mistake, I reverted it. I also squashed all commits into one.

PR #1086, opened 2023-04-14
fix: Fix refresh insight scheduler

### LINKED ISSUE #958: We don't check anymore that insights are still valid in scheduler
We used to check every night in `refresh_insights` function (in the scheduler) that insights are still valid. After the prediction/insight split, this feature has been lost.
Reintroduce it, as we start to have insights that suggest modifications already applied in production.

raphael0202:
- loading min dataset in memory took +20GB of RAM, which made the
scheduler crash. By allowing a projection (selected fields) to be passed
as parameter, we reduce memory usage to ~2GB.
- Use ServerSide cursor to avoid excessive memory consumption
- Also delete predictions associated with deleted products

fixes #958
We still need to check that the image has not been deleted.

### DIFF
--- a/robotoff/scheduler/__init__.py
+++ b/robotoff/scheduler/__init__.py
@@ -41,7 +41,7 @@ logger = get_logger(__name__)
 
 
 # Note: we do not use with_db, for atomicity is handled in annotator
-def process_insights():
+def process_insights() -> None:
     with db.connection_context():
         processed = 0
         insight: ProductInsight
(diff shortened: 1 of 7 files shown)


robotoff/scheduler/__init__.py
```diff
@@ -124,50 +124,69 @@ def refresh_insights(with_deletion: bool = False):
...
         )
-        to_update = True
+        updated_fields.append("unique_scans_n")
         insight.unique_scans_n = product.unique_scans_n
 
-    if to_update:
-        insight.save()
+    if updated_fields:
+        # Only update selected field with bulk_update and a list of fields to update
+        ProductInsight.bulk_update([insight], fields=updated_fields)
```
alexgarel, line 183: :+1:

alexgarel (approved):
LGTM

PR #1091, opened 2023-04-16
fix: always use the same queue for jobs of the same product

raphael0202:
We create as many high priority queue as they are workers, and assign each worker (whether high or low priority) to a unique queue. This way, we reduce the risk of concurrent DB write for the same product.

Also, send webhook update jobs for all projects, as we now have a MongoDB connection for all projects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,8 +118,9 @@ dl-model-categorizer:
 lauch-burst-worker:
 ifdef queues
 	${DOCKER_COMPOSE} run --rm -d --no-deps worker_low python -m robotoff run-worker ${queues} --burst
+# Only launch burst worker on low priority queue if queue is not specified
 else
-	${DOCKER_COMPOSE} run --rm -d --no-deps worker_low python -m robotoff run-worker robotoff-high robotoff-low --burst
+	${DOCKER_COMPOSE} run --rm -d --no-deps worker_low python -m robotoff run-worker robotoff-low --burst
 endif
 
 #------------#
@@ -168,26 +169,26 @@ health:
 i18n-compile:
 	@echo "🥫 Compiling translations …"
 # Note it's important to have --no-deps, to avoid launching a concurrent postgres instance
-	${DOCKER_COMPOSE} run --rm --entrypoint bash --no-deps worker_high -c "cd i18n && . compile.sh"
+	${DOCKER_COMPOSE} run --rm --entrypoint bash --no-deps worker_high_1 -c "cd i18n && . compile.sh"
 
 unit-tests:
 	@echo "🥫 Running tests …"
 	# run tests in worker to have more memory
 	# also, change project name to run in isolation
-	${DOCKER_COMPOSE_TEST} run --rm worker_high poetry run pytest --cov-report xml --cov=robotoff tests/unit
+	${DOCKER_COMPOSE_TEST} run --rm worker_high_1 poetry run pytest --cov-report xml --cov=robotoff tests/unit
 
 integration-tests:
 	@echo "🥫 Running integration tests …"
 	# run tests in worker to have more memory
 	# also, change project name to run in isolation
-	${DOCKER_COMPOSE_TEST} run --rm worker_high poetry run pytest -vv --cov-report xml --cov=robotoff --cov-append tests/integration
+	${DOCKER_COMPOSE_TEST} run --rm worker_high_1 poetry run pytest -vv --cov-report xml --cov=robotoff --cov-append tests/integration
 	( ${DOCKER_COMPOSE_TEST} down -v || true )
 
 # interactive testings
 # usage: make pytest args='test/unit/my-test.py --pdb'
 pytest: guard-args
 	@echo "🥫 Running test: ${args} …"
-	${DOCKER_COMPOSE_TEST} run --rm worker_high poetry run pytest ${args}
+	${DOCKER_COMPOSE_TEST} run --rm worker_high_1 poetry run pytest ${args}
 
 #------------#
 # Production #
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ dl-model-categorizer:
 	tar -xzvf saved_model.tar.gz --strip-component=1; \
 	rm saved_model.tar.gz
 
-lauch-burst-worker:
+launch-burst-worker:
 ifdef queues
 	${DOCKER_COMPOSE} run --rm -d --no-deps worker_low python -m robotoff run-worker ${queues} --burst
 # Only launch burst worker on low priority queue if queue is not specified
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -118,8 +118,9 @@ dl-model-categorizer:
 lauch-burst-worker:
```
alexgarel, line 118: lauch :-)

docker/dev.yml
```diff
@@ -33,18 +33,15 @@ x-robotoff-dev: &robotoff-dev
 services:
   api:
     <<: *robotoff-dev
-  worker_high:
+  worker_high_1:
```
alexgarel, line 36: you are defining a service which is completely different from worker_high.

I don't think it's what you intended to do…

robotoff/workers/queues.py
```diff
@@ -1,23 +1,58 @@
...
+from robotoff.types import ProductIdentifier
+from robotoff.utils import get_logger
 
-high_queue = Queue(WorkerQueue.robotoff_high.value, connection=redis_conn)
-low_queue = Queue(WorkerQueue.robotoff_low.value, connection=redis_conn)
+logger = get_logger(__name__)
+high_queues = [
+    Queue(f"robotoff-high-{i+1}", connection=redis_conn)
+    for i in range(settings.NUM_RQ_WORKERS)
+]
```
alexgarel, line 18: NUM_RQ_WORKERS is 4 but you have 2 services ?

robotoff/workers/queues.py
```diff
@@ -1,23 +1,58 @@
...
+    :return: the selected queue
+    """
+    if product_id is None:
+        return random.choice(high_queues)
+
+    # All product barcodes should be digits, but just make sure the function still
+    # works with non-digit barcodes
+    queue_idx = (int(product_id.barcode) if product_id.barcode.isdigit() else 0) % len(
+        high_queues
+    )
```
alexgarel, line 45: do you think % works well here ? Do we have an even distribution of last digits for products code ? Also we might want to support non numerical barcode in the future. So why not hash the barcode before using modulo ?

alexgarel (commented):
I did some tests and I think we could instead use deploy / replicas to have this functionality in a more flexible way.

If you launch replicas, you can target a specific container by using <COMPOSE_PROJECT_NAME>-<CONTAINER>-<NUMBER> (eg robotoff-high_queue-1, robotoff-high_queue-2)

You still have to know how many replica are available and I see no way to do it without passing it in an env variable.

docker/dev.yml
alexgarel: oups sorry this one, was unintentional (I did not see you modified docker-compose.yml)

robotoff/workers/queues.py
raphael0202: We have 4 services, no? `low_{1,2}` and `high_{1,2}`

robotoff/workers/queues.py
raphael0202: I thought about it, and considered it was a bit overkill to use a hashing function for this. I don't know how even the distribution is though.

raphael0202:
> I did some tests and I think we could instead use deploy / replicas to have this functionality in a more flexible way.
> If you launch replicas, you can target a specific container by using <COMPOSE_PROJECT_NAME>-- (eg robotoff-high_queue-1, robotoff-high_queue-2)
> You still have to know how many replica are available and I see no way to do it without passing it in an env variable.

How would it work exactly? I didn't find a way to provide the container name to the running container, which is required to know which queue we should listen to.

robotoff/workers/queues.py
alexgarel: yes but here it's only `robotoff-high-`, isn't it ?

robotoff/workers/queues.py
alexgarel: I would find it more robust. (a md5 or sha1 does not cost a lot)

alexgarel:
> How would it work exactly? I didn't find a way to provide the container name to the running container, which is required to know which queue we should listen to.

It's really a pity but are right :cry: 

One way we could handle this is by using a redis counter modulo the number of replica. https://redis.io/commands/incr/ … but even then I see scenario where it would not work, like stopping one of the replica and restarting it.

So it does not seems worth it…

Ok for hardcoding it then !

robotoff/workers/queues.py
alexgarel: ok, my bad, I didn't see low queues where also listening for high queues.

Why aren't all worker listening the low queue then ?

I don't really understand your pattern of naming some high / other low, while they still have a high queue to listen…

Are there some tasks that you only dedicate to worker_high_1/2 ?

docker-compose.yml
```diff
@@ -52,27 +64,27 @@ services:
...
-    command: python -m robotoff run-worker robotoff-high
-    environment: *robotoff-base-env
-    depends_on:
-      - postgres
-    mem_limit: 8g
+  worker_high_1:
+    <<: *robotoff-worker
+    container_name: ${COMPOSE_PROJECT_NAME:-robotoff}_worker_high_1
+    # Each worker (whether it's a high or low priority worker) listens to a
+    # single high priority queue
```
alexgarel, line 71: I would put that comment below on  _low_1 where it's more easy to miss the fact it's listening to a high queue.

robotoff/workers/queues.py
raphael0202: Only low priority workers are listening to low queues. The idea was that all workers handle high priority jobs first, but only a limited number of workers (=2) can handle low priority jobs, so that the server is not stressed (as we can have 1000s of low priority jobs when launched using CLI).

robotoff/workers/queues.py
raphael0202: You're right, I guess it was just laziness from my part :smile: 
I added it (+ some unit tests)

raphael0202:
> It's really a pity but are right cry
> One way we could handle this is by using a redis counter modulo the number of replica. https://redis.io/commands/incr/ … but even then I see scenario where it would not work, like stopping one of the replica and restarting it.
> So it does not seems worth it…
> Ok for hardcoding it then !

I'm not a fan of this solution, but it's the best one I came up with. And it will solve concurrency errors that stack up in Sentry (see https://openfoodfacts.sentry.io/issues/4094546464/?project=1415205&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+none%5D&referrer=issue-stream&sort=inbox&stream_index=2 or https://openfoodfacts.sentry.io/issues/4101632975/?project=1415205&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+none%5D&referrer=issue-stream&sort=inbox&stream_index=0)

PR #1099, opened 2023-04-20
chore: fix translation on fr

alexgarel:
Fix a translation issue (also fixed in crowdin).


alexgarel:
@raphael0202 it might be better to merge the crowdin automatic PR. As you like :-)

PR #1101, opened 2023-04-21
fix: fix error raised when releasing expired lock

raphael0202:
See https://openfoodfacts.sentry.io/issues/4114245248/?alert_rule_id=662444&alert_timestamp=1682067470955&alert_type=email&environment=prod&project=1415205&referrer=alert_email

Job took 2 minutes, and the lock expires after 1 min. We shouldn't throw an error when trying to release a lock that expired.

### DIFF
--- a/robotoff/redis.py
+++ b/robotoff/redis.py
@@ -2,10 +2,15 @@ from typing import Optional
 
 from redis import Redis
 from redis_lock import Lock as BaseLock
+from redis_lock import NotAcquired
 
 from robotoff import settings
+from robotoff.utils import get_logger
 
 redis_conn = Redis(host=settings.REDIS_HOST)
+
+
+logger = get_logger(__name__)
 
 
 class LockedResourceException(Exception):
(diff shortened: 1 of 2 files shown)


alexgarel (approved):
Shan't we also make lock stand for a longer time. (I mean the time to live is only for pathological case where lock is not released).

robotoff/redis.py
```diff
@@ -39,4 +44,10 @@ def __enter__(self):
...
     def __exit__(self, *args, **kwargs):
         if self._enabled:
-            self.release()
+            try:
+                self.release()
+            except NotAcquired:
+                logger.info(
+                    "releasing lock failed, lock %s does not exist or expired",
+                    self._name,
+                )
```
alexgarel, line 53: I think it's worth tracking those kind of errors (they should not happen too much, otherwise we really have to extend the time to live). So I would, at least, use logger.warning

raphael0202: Good point, I've increased it to 5 min + replaced info by warning.

PR #1151, opened 2023-07-24
refactor: add off sdk

raphael0202:
Use OFF SDK to replace part of Robotoff codebase.


alexgarel:
Kudos for this move @raphael0202

PR #1167, opened 2023-08-11
fix: add robots.txt

raphael0202:
### What

Disallow completely indexation: otherwise, web crawlers send millions of requests to Robotoff (420k requests/day by Google alone)


teolemon (approved):
Say whaaaaat ?

teolemon:
What about sending blank JS files ? @raphael0202

raphael0202:
> What about sending blank JS files ?

This should be sufficient to prevent crawlers from accessing the website

PR #1184, opened 2023-08-28
feat: Added Polish brands, stores and labels

ArturLange:
### What
- I added regexes to stores and brands popular in Poland and links to popular Polish labels


data/ocr/label_flashtext.txt
```diff
@@ -114,6 +115,7 @@ en:sustainable-palm-oil||sustainable palm oil
 en:sustainable-seafood-msc||www.msc.org
 en:sustainable-seafood-msc||pêche durable msc
 en:utz-certified||utz certified
+en:vegan||suitable for vegans
```
raphael0202, line 118: As vegan is already there you can just add "vegans" as a pattern.

data/ocr/store_regex.txt
```diff
@@ -1,22 +1,49 @@
 7-eleven
+Action
```
raphael0202, line 2: Hum it will introduce many false positive, it's quite a common word.

data/ocr/store_regex.txt
```diff
@@ -1,22 +1,49 @@
...
 AhorraMás
 AhorraMás||ahorramas
 Albertsons
 Albert Heijn
 Alcampo
 Aldi
+Aldi||asia green garden
+Aldi||balta mare
+Aldi||biscotto
```
raphael0202, line 11: Same thing about false positive maybe.

raphael0202 (changes requested):

data/ocr/brand_logo_annotation.txt
```diff
@@ -257,6 +257,7 @@ toupargel||Toupargel
 tropicana orange juice||Tropicana
 tropicana||Tropicana
 Tropicana||Tropicana
+TYMBARK||Tymbark
```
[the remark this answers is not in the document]
ArturLange: You're right, that wasn't obvious to me, I thought it's supposed to be the way it's written on the packaging.
In that case I shouldn't add those without knowledge about Google Cloud Vision, is that right?

data/ocr/store_regex.txt
teolemon, line 11:
```suggestion
```

data/ocr/store_regex.txt
teolemon, line 2:
```suggestion
```

data/ocr/brand_logo_annotation.txt
raphael0202, line 260:
```suggestion
```

data/ocr/label_flashtext.txt
raphael0202, line 118:
```suggestion
en:vegan||vegans
```

ArturLange:
@raphael0202 I removed **Netto** from `store_regex.txt` after your comments - it's also a common word on packaging.
Is it ok if I just leave the various own brands of **Netto**? 
```
Netto||bergio
Netto||corsarro
Netto||happy vege
Netto||miletto
Netto||paradiso
Netto||quallio
Netto||rajski sad
Netto||sztuka mięsa
Netto||toremo
```

raphael0202:
Stores are applied automatically, that's the issue... I'm not super fan of this store system we put in place, I would much rather deduce the store from the brand in Robotoff (see https://github.com/openfoodfacts/robotoff/discussions/1225).
If you don't mind I would much rather not include additional stores in this file, as it takes time to clean the DB once false positive are introduced).

data/ocr/store_regex.txt
```diff
@@ -242,6 +276,15 @@ Mega Image
 Migros
 Monoprix
 Morrisson's
+Netto||bergio
+Netto||corsarro
+Netto||happy vege
+Netto||miletto
+Netto||paradiso
```
teolemon, line 283:
```suggestion
```

raphael0202:
As we don't want to rely anymore on manual detection to detect store/brands (due to risks of false positive), I'm closing this PR.

PR #1246, opened 2023-08-29
feat: apply automatically nutrition images insights

raphael0202:
Predictions are very good:
https://hunger.openfoodfacts.org/questions?type=nutrition_image


alexgarel (commented):
@teolemon I think you should be the one approving this.

teolemon:
It seems that it's been on since May, and has produced only 1K insights. What is the point of auto-application ?

raphael0202:
It's 10 products fixed per day, and the risk is low :)

PR #1267, opened 2023-10-16
fix: update pymongo

raphael0202:
After incident of the weekend: https://github.com/openfoodfacts/openfoodfacts-infrastructure/pull/265


raphael0202:
@stephanegigandet From what I understand, multiple connections should only happen in a multi-threaded setting (https://pymongo.readthedocs.io/en/stable/faq.html#how-does-connection-pooling-work-in-pymongo), which never happens for us: we instanciate the MongoClient after thread fork in all robotoff services.

PR #1275, opened 2023-10-25
feat: add DB migration library

raphael0202:
- add migrate_peewee library to handle DB migrations
- try with a new `product_insight.bounding_box` field

### DIFF
--- a/gunicorn.py
+++ b/gunicorn.py
@@ -12,6 +12,6 @@ timeout = 60
 
 def on_starting(server):
     """Gunicorn server hook."""
-    with models.db:
-        models.db.execute_sql("CREATE SCHEMA IF NOT EXISTS embedding;")
-        models.db.create_tables(models.MODELS, safe=True)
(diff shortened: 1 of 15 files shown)


gunicorn.py
```diff
@@ -12,6 +12,6 @@
 
 def on_starting(server):
     """Gunicorn server hook."""
-    with models.db:
-        models.db.execute_sql("CREATE SCHEMA IF NOT EXISTS embedding;")
-        models.db.create_tables(models.MODELS, safe=True)
+    # Perform migrations
+    with models.db.connection_context():
+        models.run_migration()
```
alexgarel, line 17: Normally it's not considered a good idea to run migration on startup, at least in production. 
It is considered a better strategy to explicitly run them at deployment (or project setup for dev).

alexgarel (approved):
Cool to have migrations.

Just added an opinionated comment :-)

PR #1279, opened 2023-10-26
fix: more stores

teolemon:
### What
- fix: more stores

### DIFF
--- a/data/ocr/store_regex.txt
+++ b/data/ocr/store_regex.txt
@@ -11,7 +11,6 @@ Aldi||gutbio
 Aldi||milsani
 Aldi||sprengel
 Aldi||liveGfree
-Aldi||Simply Nature
 Aldi||Little Journey
 Aldi||Winking Owl
 Aldi||Ambiano
--- a/data/ocr/store_regex.txt
+++ b/data/ocr/store_regex.txt
@@ -10,6 +10,15 @@ Aldi||gold[aä]hren
 Aldi||gutbio
 Aldi||milsani
 Aldi||sprengel
+Aldi||liveGfree
+Aldi||Simply Nature
+Aldi||Specially Selected
+Aldi||Little Journey
+Aldi||Winking Owl
+Aldi||Ambiano
+Aldi||Barissimo
+Aldi||Friendly Farms
+Aldi||Millville
 Alimerka
 Alnatura
 Alteza
@@ -42,6 +51,7 @@ Carrefour
 Carrefour||grand jury
 Casino
 Colruyt
+Conad
 Costco
 Coviran
 Delhaize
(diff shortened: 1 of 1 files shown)


data/ocr/store_regex.txt
```diff
@@ -42,6 +51,7 @@ Carrefour
 Carrefour||grand jury
 Casino
 Colruyt
+Conad
```
raphael0202, line 54: Conad is already there (hence the failing test)

data/ocr/store_regex.txt
```diff
@@ -10,6 +10,15 @@ Aldi||gold[aä]hren
 Aldi||gutbio
 Aldi||milsani
 Aldi||sprengel
+Aldi||liveGfree
+Aldi||Simply Nature
+Aldi||Specially Selected
```
raphael0202, line 15: Specially selected may be detected on many products

data/ocr/store_regex.txt
```diff
@@ -10,6 +10,15 @@ Aldi||gold[aä]hren
...
+Aldi||Little Journey
```
raphael0202, line 16: Same thing for little journey or friendly farms

teolemon: those combinations of words are unlikely to happen ? I'm removing simply nature

data/ocr/store_regex.txt
```diff
@@ -10,6 +10,14 @@ Aldi||gold[aä]hren
 Aldi||gutbio
 Aldi||milsani
 Aldi||sprengel
+Aldi||liveGfree
+Aldi||Simply Nature
```
teolemon, line 14:
```suggestion
```

data/ocr/store_regex.txt
raphael0202: Well technically they can. The thing is, it is autoapplied here. We should either check that it doesn't occur in the dataset, or have a way to have validation from a human.

data/ocr/store_regex.txt
```diff
@@ -10,6 +10,13 @@ Aldi||gold[aä]hren
 Aldi||gutbio
 Aldi||milsani
 Aldi||sprengel
+Aldi||liveGfree
+Aldi||Little Journey
```
teolemon, line 14:
```suggestion
```

data/ocr/store_regex.txt
```diff
@@ -10,6 +10,13 @@ Aldi||gold[aä]hren
...
+Aldi||Winking Owl
+Aldi||Ambiano
+Aldi||Barissimo
+Aldi||Friendly Farms
```
teolemon, line 18:
```suggestion
```

PR #1296, opened 2023-12-04
refactor: upgrade to falcon 3.X

raphael0202:
- switch to falcon 3.1. It has better exception handling, and supports more recent versions of Python officially.
- add auto-reload support for gunicorn, and enable it during dev by default


alexgarel:
:clap:

PR #1311, opened 2024-02-07
Update label_logos.json

alexfauquette:
I noticed logos for nutriscore were broken. 
After a quick search, All are good except those two logo I did not found

- en:european-vegetarian-union-vegetarian
- en:european-vegetarian-union-vegan


teolemon (approved):
LGTM

raphael0202:
Thank you for the PR!

PR #1366, opened 2024-07-16
fix: Use an allowlist for the crop endpoint

VaiTon:
### What
- Use an allowlist to limit the domains on which the crop operation is allowed.
- It is a bad security practice to allow any url there


.env
```diff
@@ -99,4 +99,6 @@ LOG_LEVEL=DEBUG
 # Use 16G in production
 ROBOTOFF_POSTGRES_SHARED_BUFFERS=1GB
 # Use 1G in production
-ROBOTOFF_POSTGRES_WORK_MEM=64MB
\ No newline at end of file
+ROBOTOFF_POSTGRES_WORK_MEM=64MB
+
+CROP_ALLOWED_DOMAINS=static.openfoodfacts.org
```
raphael0202, line 104:
```suggestion
CROP_ALLOWED_DOMAINS=static.openfoodfacts.org,openfoodfacts-images.s3.eu-west-3.amazonaws.com,images.openfoodfacts.org
```

raphael0202:
Thank you for your PR!
I just added a suggestion to allow more domain (especially our S3 backup)

raphael0202 (approved):

PR #1377, opened 2024-07-19
fix: avoid failing all metrics for a single failure

alexgarel:
Add try / except so that a failure on one facet does not block every other facets.


raphael0202 (approved):
Great, thank you!

PR #1436, opened 2024-10-22
feat: :zap: Add CLI command to convert and push JSONL to Huggingface

jeremyarancio:
### What

To make the Open Food Facts database easier to share, we decided to upload it on a regular basis to the Hugging Face platform.
This PR adds a CLI commands to clean, convert to parquet and push to HF the JSONL dump.

The dataset is already deployed in the [Open Food Facts HF repo](https://huggingface.co/datasets/openfoodfacts/product-database)


raphael0202:
Pinging @stephanegigandet and @alexgarel to discuss the fields to include/remove.

A few random remarks:

- in my opinion, we should include the tagged version of a field, and not the original string as provided by the user. `countries` -> `countries_tags`, `states` -> `states_tags`, `categories` -> `categories_tags`,...
- `images` is a string in the output parquet file, it should be a dict
- we should reformat the language-specific fields such as `ingredients_text_fr`, `ingredients_text_en` to have a single `ingredients_text` field like: `{"main": "ING", "fr": "ING_FR", "en": ING_EN"}`. `main` here is the ingredient field for the main language (`ingredients_text`)

jeremyarancio:
@raphael0202 
To enable pushing to HF, we need to add `HF_TOKEN` to the env variables.

raphael0202 (approved):
It looks good to me, thank you!

PR #1518, opened 2025-01-31
fix: making changes in importer.py and annotate.py

prasanna7codes:
### What
- 
-Implemented the Image Orientation Auto-Rotation feature.
ImageOrientationImporter class parses the prediction.data JSON to determine the required rotation based on word orientation.
-ImageOrientationAnnotator class rotates the image if needed based on the rotation angle from the importer.
Integrated logic to check if the predicted orientation is not "up" and confidence is >= 95% before rotating.

### Fixes bug(s)
- 
- #Fixed issue with incorrectly rotated images from contributors.
#Fixed issue where images were not rotated automatically if the predicted orientation was incorrect.

### Part of 
-

### DIFF
--- a/robotoff/insights/annotate.py
+++ b/robotoff/insights/annotate.py
@@ -116,6 +116,39 @@ CANNOT_VOTE_RESULT = AnnotationResult(
     description="The voting mechanism is not compatible with this insight type, please authenticate.",
 )
 
+from PIL import Image
+from importer import ImageOrientationImporter
+
+class ImageOrientationAnnotator:
+    @staticmethod
+    def rotate_image(image_path, rotation_angle):
+        """
+        Rotate the image by the specified angle.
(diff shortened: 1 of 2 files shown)


alexgarel:
@prasanna7codes thanks a lot for this contribution.

Sorry because it's a long time since I worked on robotoff, and main robotoff contributor is AFK for some time !

That said I'm not sure the way you wrote it is not the way robotoff work.

The importer should generate an insight, which if auto applied, will generate the annotation.
See https://openfoodfacts.github.io/robotoff/explanations/predictions/ and https://openfoodfacts.github.io/robotoff/how-to-guides/add-predictor/

Please take a look how it works for the others.

Also a test would be appreciated if you can.

You can come on the slack.openfoodfacts.org in #robotoff to discuss it if needed.

PR #1585, opened 2025-05-23
feat: Use FaceAnnotations

areebahmeddd:
WIP

### DIFF
--- a/robotoff/cli/ocr.py
+++ b/robotoff/cli/ocr.py
@@ -16,7 +16,12 @@ def run_ocr_on_image_batch(base64_images: List[str], api_key: str) -> requests.R
         json={
             "requests": [
                 {
-                    "features": [{"type": "TEXT_DETECTION"}],
+                    "features": [
+                        {"type": "TEXT_DETECTION"},
+                        {"type": "FACE_DETECTION"},
+                        {"type": "LABEL_DETECTION"},
+                        {"type": "SAFE_SEARCH_DETECTION"},
+                    ],
                     "image": {"content": base64_image},
                 }
                 for base64_image in base64_images


raphael0202 (changes requested):
LGTM, except:

- code added to legacy script
- minor remarks
- we need to account for the new `type=face_annotation` in `notifier.py` and set `reason=human` so that it is sent correctly to Nutripatrol

tests/unit/prediction/ocr/test_image_flag.py
```diff
@@ -0,0 +1,153 @@
...
+from robotoff.prediction.ocr.image_flag import flag_image
+from robotoff.types import PredictionType
+
+
+@pytest.fixture
+def face_detection_ocr_result():
+    test_data_url = "https://raw.githubusercontent.com/openfoodfacts/test-data/main/robotoff/tests/unit/ocr/face_detection.json"
+    response = requests.get(test_data_url)
+    response.raise_for_status()
+    return OCRResult.from_json(response.json())
```
raphael0202, line 14: you can directly use `OCRResult.from_url(test_data_url)` here

tests/unit/prediction/ocr/test_image_flag.py
```diff
@@ -0,0 +1,153 @@
...
+
+    face_predictions = [
+        pred
+        for pred in predictions
+        if pred.type == PredictionType.image_flag
+        and pred.data.get("type") == "face_annotation"
+    ]
+
+    assert face_predictions
+    assert len(face_predictions) == 1
```
raphael0202, line 30: No need to have the first assert, as we already know it's a list, both assert are redundant

tests/unit/prediction/ocr/test_image_flag.py
```diff
@@ -0,0 +1,153 @@
...
+    face_predictions = [
+        pred
+        for pred in predictions
+        if pred.type == PredictionType.image_flag
+        and pred.data.get("type") == "label_annotation"
+        and pred.data.get("label") == "face"
+    ]
+
+    assert face_predictions, "No face label predictions were detected"
+    assert len(face_predictions) == 1, "Expected exactly one face label prediction"
```
raphael0202, line 56: Same remark as above

tests/unit/prediction/ocr/test_image_flag.py
```diff
@@ -0,0 +1,153 @@
...
+    assert face_predictions[0].data["likelihood"] == 0.8
```
raphael0202, line 57: You can directly check all values of the `Prediction`, doing a `assert face_predictions[0] == Prediction(...)`. In my opinion, it is good to perform a full check at least for one test case.

tests/unit/prediction/ocr/test_image_flag.py
```diff
@@ -0,0 +1,153 @@
...
+        for pred in predictions
+        if pred.type == PredictionType.image_flag
+        and pred.data.get("type") == "safe_search_annotation"
+        and pred.data.get("label") == "adult"
+    ]
+
+    assert safe_search_predictions, "No safe search predictions were detected"
+    assert (
+        len(safe_search_predictions) == 1
+    ), "Expected exactly one safe search prediction"
```
raphael0202, line 89: Same remark as above

raphael0202 (changes requested):
I added a few small remarks on test, but otherwise it's all good!

raphael0202 (approved):

PR #1615, opened 2025-06-04
fix: Add retry logic for redis

areebahmeddd:
**Fixes** #1612

**Problem**
Update listener crashes when Redis connections drop, causing service termination.

**Solution**

* Added `backoff` with `@on_exception` on `run_update_listener()` for retry handling
* Configured exponential backoff: 10 retries, 1-minute timeout, random jitter

### DIFF
--- a/robotoff/workers/update_listener.py
+++ b/robotoff/workers/update_listener.py
@@ -1,8 +1,11 @@
+import time
+
 from openfoodfacts import Environment, Flavor
 from openfoodfacts.images import generate_image_url, generate_json_ocr_url
 from openfoodfacts.redis import RedisUpdate
 from openfoodfacts.redis import UpdateListener as BaseUpdateListener
 from redis import Redis
+from redis.exceptions import ConnectionError
 
 from robotoff import settings
 from robotoff.types import ProductIdentifier, ServerType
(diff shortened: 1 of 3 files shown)


raphael0202 (changes requested):
Thank you for your PR!
I think we should use an external library such as https://github.com/litl/backoff instead of coding it from scratch, as it's less error-prone (and unit-tested).

robotoff/workers/update_listener.py
```diff
@@ -102,16 +106,43 @@ def process_redis_update(self, redis_update: RedisUpdate):
...
+                redis_stream_name=settings.REDIS_STREAM_NAME,
+                redis_latest_id_key=settings.REDIS_LATEST_ID_KEY,
+            )
+            update_listener.run()
+        except Exception as e:
+            logger.critical(
+                "Unexpected error in update listener: %s", str(e), exc_info=True
+            )
+            # for non-connection errors, wait a bit before retrying
+            time.sleep(5)
```
raphael0202: No need to have this block, as we're catching here any error, including ConnectionError. It's good to use logger.critical to log it in Sentry though, so we can keep the `while True` loop.
```suggestion
```

robotoff/workers/update_listener.py
```diff
@@ -102,16 +106,43 @@ def process_redis_update(self, redis_update: RedisUpdate):
                 )
 
 
+@backoff.on_exception(
+    backoff.expo,
+    ConnectionError,
+    max_tries=10,
+    max_time=300,  # 5 minutes
```
raphael0202, line 111: I would suggest to:
- remove `max_tries` and `max_time` completely. We get all our events from Redis, we want to try again until Redis is not down.
- Limit the maximum time to wait to 60s using `max_value`: it's not a web server with a public API, it's okay to give it a try every minute. 

```suggestion
    max_value=60,  # we wait at most 60 seconds between retries
```

robotoff/workers/update_listener.py
```diff
@@ -102,16 +106,43 @@ def process_redis_update(self, redis_update: RedisUpdate):
...
+    jitter=backoff.random_jitter,
+    on_backoff=lambda details: logger.error(
+        "Redis connection error (attempt %d/%d): %s. Retrying in %.1f seconds...",
+        details["tries"],
+        10,
+        details["exception"],
+        details["wait"],
+    ),
```
raphael0202, line 119: This has to be updated given the suggestion above.

```suggestion
    on_backoff=lambda details: logger.error(
        "Redis connection error (attempt %d): %s. Retrying in %.1f seconds...",
        details["tries"],
        details["exception"],
        details["wait"],
    ),
```

raphael0202 (changes requested):
I added a few minor remarks, but otherwise it works well!

raphael0202 (approved):
I tested it locally and it works well, thank you!

PR #1616, opened 2025-06-04
fix: Add fallback mechanism for database migration

areebahmeddd:
Added SQL fallback that inserts missing migration records using WHERE NOT EXISTS when make migrate-db fails.
Fixes the failed deployment check [ [ref ](https://github.com/openfoodfacts/robotoff/actions/runs/15438459127/job/43450340913)]


raphael0202:
As discussed directly with @areebahmeddd, I'm closing this PR.

PR #1669, opened 2025-07-04
chore: fixed some issue and not major things

JagjeevanAK:
### What
I saw some wired problems in api doc generation so this PR fixes that issues.


JagjeevanAK:
@raphael0202 I have fixed that issues as it is not an major PR so I will be merging it feel free to suggest changes to it and I will make another PR to add that changes.

PR #1676, opened 2025-07-10
feat: add bearer token to nutripatrol auth

Valimp:
## What

Add a bearer token to the image moderation notifier


raphael0202 (changes requested):
Thank you! I added a few comments

raphael0202 (approved):
Thank you!

PR #1700, opened 2025-08-12
chore: remove legacy labels.txt files

raphael0202:
We now specify the labels directly in Robotoff.


teolemon:
https://github.com/openfoodfacts/robotoff/blob/2cc5d9e570b6fcb6433d26e1c0da33c7685a3fef/robotoff/workers/tasks/import_image.py#L436

Issue #1708, opened 2025-08-14
Use openCV to open images sent to ML models instead of Pillow


raphael0202:
openCV is **much, much** faster:

```
image shape: (1200, 607, 3)
cv2.imdecode time: 0.010s
Image.open time: 0.028s
Image.open time (including np conversion): 0.348s
---
image shape: (400, 333, 3)
cv2.imdecode time: 0.002s
Image.open time: 0.002s
Image.open time (including np conversion): 0.113s
---
image shape: (4096, 3072, 3)
cv2.imdecode time: 0.206s
Image.open time: 0.247s
Image.open time (including np conversion): 5.123s
```

Script:

```python
import io
import time

import cv2
import numpy as np
import requests
from PIL import Image

for image_url in (
    "https://world.openfoodfacts.org/images/products/281/304/702/0270/1.jpg",
    "https://images.openfoodfacts.org/images/products/210/010/055/9939/2.jpg",
    "https://prices.openfoodfacts.org/img/0100/NLhwECq0fH.webp",
):
    content_bytes = requests.get(image_url).content

    start_time = time.monotonic()
    image = cv2.imdecode(
        np.frombuffer(content_bytes, dtype=np.uint8), cv2.IMREAD_COLOR_RGB
    )
    print(f"image shape: {image.shape}")
    print(f"cv2.imdecode time: {time.monotonic() - start_time:.3f}s")

    start_time = time.monotonic()
    image = Image.open(io.BytesIO(content_bytes))
    image.load()  # Ensure the image is fully loaded
    print(f"Image.open time: {time.monotonic() - start_time:.3f}s")
    np.array(image.getdata()).reshape((image.height, image.width, 3))
    print(
        f"Image.open time (including np conversion): {time.monotonic() - start_time:.3f}s"
    )
``

teolemon:
😲

Savvythelegend:
Hi @raphael0202 @teolemon

I’ve reviewed the code and it seems the main image loading occurs in `robotoff/utils/image.py` `(get_image_from_url)` and is then used in `robotoff/robotoff/images.py`.

I can implement a switch to use OpenCV (cv2.imdecode) instead of Pillow for ML image loading, ensuring the output stays in RGB for consistency with downstream processing. This should improve the pipeline’s speed significantly while keeping compatibility with functions like `generate_image_fingerprint` that still rely on Pillow.

I’d be happy to submit a PR for this change. Could you confirm if it’s okay to proceed?

[...]

PR #1734, opened 2025-08-29
feat: enhance _generate_image_embeddings to accept PIL and NumPy images

tsvlgd:
### What
- Enhance `_generate_image_embeddings` in `keras_category_classifier_3_0` to accept both `PIL.Image` and `NumPy` arrays.
- Convert `PIL.Image` to `NumPy` before sending to Triton for inference.
- Added logging of inference time for performance monitoring.
- Updated type hints and docstrings for clarity.

### Fixes bug(s)
- N/A — this is a performance and type-handling enhancement, not a bug fix.

### Part of 
- #1708 — Use OpenCV / NumPy arrays for ML image inputs instead of relying solely on Pillow

### DIFF
--- a/robotoff/prediction/category/neural/keras_category_classifier_3_0/__init__.py
+++ b/robotoff/prediction/category/neural/keras_category_classifier_3_0/__init__.py
@@ -2,6 +2,7 @@ import logging
 import time
 import typing
 from typing import Literal, Union
+
 import numpy as np
 from openfoodfacts.ocr import OCRResult
 from PIL import Image


raphael0202 (changes requested):
You also need to download images as numpy array here: https://github.com/openfoodfacts/robotoff/blob/02054cc8bc94647a64d327e612e47cfc8e36bae6/robotoff/prediction/category/neural/keras_category_classifier_3_0/__init__.py#L150

PR #1897, opened 2026-05-25
docs: Add details to the returned fields related to images

alexfauquette:
The first commit is the actual data addition. The other ones are suggestion of improvements from AI review

## Stuff I've learned thanks to AI review

From the spec, `schema` is for to defined objects used in input/output. Whereas `responses` and `parameters` are more specific (only input or only output)
https://spec.openapis.org/oas/v3.2.0.html#components-object

> Any sibling elements of a $ref are ignored. This is because $ref works by replacing itself and everything on its level with the definition it is pointing at.

https://swagger.io/docs/specification/v3_0/using-ref/

 ## DOcs deployement

The docs deployement got broken because of the pygments update #1882 

Looks like it does not support anymore the codeblocks nested into lists

### DIFF
--- a/docs/references/api.yml
+++ b/docs/references/api.yml
@@ -907,7 +907,7 @@ paths:
                     - type
                     - value
                 server_type:
-                  $ref: "#/components/parameters/server_type"
+                  $ref: "#/components/schemas/ServerType"
               required:
               - annotations
       responses:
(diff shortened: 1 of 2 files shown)


docs/references/api.yml
```diff
@@ -145,6 +145,8 @@ paths:
                   count:
                     type: integer
                     description: The total number of questions that meet the provided criteria
+                  # This is an array of tuple. To properly document it we need OpenAPIv3.1
+                  # https://stackoverflow.com/questions/57464633/how-to-define-a-json-array-with-concrete-item-definition-for-every-index-i-e-a
```
alexfauquette:
```suggestion
```

Issue #1095, opened 2023-04-18
Take into account obsolete collection


alexgarel:
On product opener side We now move obsolete products to their own collection (see https://github.com/openfoodfacts/openfoodfacts-server/pull/8277)

So on robotoff side, when we look for a product in mongodb we should check for both collection: the regular collection (`products`) and the obsolete one (`products_obsolete`).

raphael0202:
We also check every night that insights are still valid using product dump, it's sure that the obsolete products are still in the JSONL product dump?

alexgarel:
It's sure that the obsolete products are still in the JSONL product dump?

No they are not (yet).

That said @stephanegigandet thinks obsolete products are less than 5 000 so it's not that a priority.

PR #1934, opened 2026-09-10
fix(products): fall back to obsolete product collection

saranyaelavarthi:
Product Opener moves obsolete products into `products_obsolete`, but Robotoff only looks in `products` when fetching a product by ID. An existing obsolete product is therefore reported as missing.

This change makes `DBProductStore.get_product()` fall back to `products_obsolete` when the regular collection returns `None`. Both lookups use the same `_id` and field projection, and products from either collection pass through the existing legacy-image conversion. A match in the regular collection avoids the extra query.

Regression coverage checks both collections, projected fields, image conversion, missing products, disabled MongoDB access, and organization-prefixed producer-platform IDs. The new fallback cases failed against the original implementation.

Related to #1095. This PR covers direct MongoDB product lookups; the JSONL-dump/nightly-cleanup concern discussed in the issue remains separate.

### Validation

On Python 3.12:

- `pytest tests/unit/test_products.py -q` — 26 passed.
- `pytest tests/unit/workers/tasks/test_import_image.py tests/unit/workers/tasks/test_product_updated.py tests/unit/insights/test_importer.py -q` — 190 passed.
- `ruff check robotoff/products.py tests/unit/test_products.py` — passed.
- `ruff format --check robotoff/products.py tests/unit/test_products.py` — passed.
- `mypy --follow-imports=silent robotoff/products.py` — passed.
- `git diff --check` — passed.

The full Docker/integration suite was not run locally; Docker is unavailable in this environment.

### DIFF
--- a/robotoff/products.py
+++ b/robotoff/products.py
@@ -576,6 +576,8 @@ class DBProductStore(ProductStore):
     ) -> JSONType | None:
         """Fetch a product from the MongoDB.
 
+        Search regular products first, then obsolete products if not found.
+
         :param product_id: identifier of the product to fetch
         :param projection: list of fields to retrieve, if not provided all fields
             are queried
@@ -589,6 +591,10 @@ class DBProductStore(ProductStore):
         # barcode for pro platform, which is also the case for
         # `product_id.barcode`
         product = self.collection.find_one({"_id": product_id.barcode}, projection)
+        if product is None:
+            product = self.db.products_obsolete.find_one(
+                {"_id": product_id.barcode}, projection
+            )
 
         # Convert the `images` field to the legacy schema, until the migration
         # is done. Once it's done, we can upgrade all Robotoff code to use the new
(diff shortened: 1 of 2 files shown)


TaciteOFF (approved):
The MongoDB fallback looks good. I tested it locally.
The API correctly retrieves images for obsolete products, and the regular collection still takes priority when both contain the product.
I also confirmed the remaining JSONL limitation: if an obsolete product is missing from the dump, nightly cleanup deletes its older predictions and unannotated insights, even though the product still exists in MongoDB. The API then returns no_questions.

TaciteOFF:
We should keep #1095 open to track the remaining cleanup work.

saranyaelavarthi:
@TaciteOFF Thanks for testing and approving! Agreed—we should keep #1095 open to track the remaining JSONL/nightly-cleanup issue. This PR covers the MongoDB lookup fallback.

Issue #1535, opened 2025-03-28
Add support for xx:Brands, needed by Hunger Games (cc @alexfauquette ) because of brands taxonomization


teolemon:
### What
- Add support for xx:Brands, needed by Hunger Games (cc @alexfauquette ) because of brands taxonomization
- Alternative: tweak Knowledge Panel for Facets to return something Hunger Games and Robotoff can digest to filter their views.

![Image](https://github.com/user-attachments/assets/c337a9f7-3022-4dce-b0cd-8f0ce8987b0f)

https://hunger.openfoodfacts.org/questions?country=en%3Aworld&type=brand&value_tag=xx%3Anestle

@stephanegigandet @alexgarel @raphael0202

stephanegigandet:
> It looks like we only use the prefix in the taxonomy, and that the tags in `brands_tags` are not prefixed with `xx`, ex:
> 
> https://world.openfoodfacts.org/api/v2/product/7611612142571?fields=brands,brands_tags
> 

@raphael0202 Sorry for the late reply. In order not to break existing apps, we remove the xx: on API queries with a low version.
Try https://world.openfoodfacts.org/api/v3.4/product/7611612142571?fields=brands,brands_tags 

See https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/api/ref-api-and-product-schema-change-log.md#2025-03-12---product-version-1001---api-version-32---removed-ingredients_hierarchy-added-schema_version-made-brands-a-taxonomized-field

raphael0202:
Indeed! Thank you for the heads up

PR #1936, opened 2026-09-13
fix(insights): accept xx-prefixed brand value filters

saranyaelavarthi:
Related to #1535. Proposed Robotoff-side normalization; design agreement and PostgreSQL validation are required before merging.

This revision normalizes brand prediction and insight value_tag values to the xx: form, adds an in-place migration, and accepts legacy brand filters for predictions and insights. After migration, responses and grouped counts use canonical stored tags. Brand validation and duplicate detection handle both input formats.

Validation: 194 unit tests passed; Ruff, formatting, mypy, and diff checks passed. Eighteen relevant PostgreSQL cases collected, but execution is blocked by the missing local database service.

Deployment requires coordinating writers with migration and validating against staging data. See docs/how-to-guides/brand-tag-normalization.md.


TaciteOFF (changes requested):
Thanks for working on this — the problem is real. Brands are now a taxonomized field, so the canonical tag is xx:nestle. The API still strips the xx: prefix on older versions to avoid breaking existing apps (v2 and v3.1 return monte-castello, v3.4 returns xx:monte-castello), but clients on recent API versions get the prefixed form, while Robotoff stores the unprefixed one. So Hunger Games gets no results.

I have a few concerns about the approach though.

1. Where the fix belongs is still open. In #1535, nobody has confirmed that Robotoff is the right place. It would be good to settle that in the issue before merging.

2. This fixes reading, not the data. The mismatch comes from what we store. Normalizing brand value_tag at insert time (plus a one-off migration) would fix the problem everywhere, instead of in one query.

3. Some paths are still broken, which follows from point 2:

- get_predictions() (core.py:295) still uses exact matching, so /api/v1/predictions?value_tag=xx:nestle returns nothing.
- group_by_value_tag=True (core.py:214, used by the insight count endpoint) will now return two separate groups, nestle and xx:nestle, for the same brand. That's a new inconsistency.
- Responses still contain the stored tag, so a client that filters again on value_tag in the frontend will see a mismatch.

4. Smaller things:

- "brand" is hardcoded; InsightType.brand.name would be safer.
- The clause is redundant: for value_tag="nestle", value_tag == 'nestle' is already covered by the IN (...).
- tests/unit/test_core.py monkeypatches ProductInsight with a fake SQLite model. That doesn't match the real schema or PostgreSQL behaviour, and it overlaps with the integration test. This repo usually covers this kind of thing with integration tests.
- You mention the new PostgreSQL tests were collected but not run locally — worth confirming once CI is green.

saranyaelavarthi:
@TaciteOFF Thanks for the detailed review. You’re right that the original query-only fix left gaps.

I’ve pushed a proposed revision that normalizes brand tags before prediction insertion and insight creation, with a migration for existing records. It also handles legacy prediction filters and preserves blacklist, barcode-range, and duplicate checks. I removed the SQLite stand-in tests and added PostgreSQL coverage for migration, responses, grouped counts, and vote preservation.

The 194 relevant unit tests and lint/type checks pass. PostgreSQL tests collect successfully, but I couldn’t execute them locally because no database service is available. The existing green PR checks don’t run these tests.

I agree we should confirm the intended approach in #1535 before merging. Would normalization in Robotoff be the preferred direction?

